PR #8, opened 2021-11-04
Init build/test scripts

andrewballantyne:
The config @florkbr is using to setup the OpenShift Bot to get `/lgtm` and `/approve` (among other things) into the HAC repos requires a couple of scripts. This PR is to add those so the bot can execute successfully on your repo.

Please https://github.com/openshift/release/pull/23230 for the release config powering this. 

Please merge this PR when you get a chance and the above PR probably will need a `/retest`. Hopefully it will pass at that point. If it does, the owners in this repo should have power over merging it with normal `/approve`-esk mechanics.

Once both of these PRs are merged. You should be good to make use of the OpenShift Bot as we do in OpenShift Console.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+set -euo pipefail
+
+echo Perform a build!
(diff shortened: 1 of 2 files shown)


jerolimov (approved):
/lgtm
/approve

PR #10, opened 2021-11-07
Add catalog components to create a generic catalog

divyanshiGupta:
Story: https://issues.redhat.com/browse/HAC-129


rohitkrai03:
Don't think we should bring in all the styles from the console.

src/shared/components/catalog/CatalogController.tsx
```diff
@@ -0,0 +1,205 @@
...
+      crumbs.push({
+        name: title,
+        path: `${pathname}?${CatalogQueryParams.TYPE}=${type}`,
+      });
+    }
+
+    return crumbs;
+  }, [pathname, t, queryParams, title, type]);
+
+  // const selectedItem = React.useMemo(() => {
```
rohitkrai03, line 112: Why is this commented out?

src/shared/components/catalog/CatalogController.tsx
```diff
@@ -0,0 +1,205 @@
...
+                items={catalogItems}
+                categories={categories}
+                filters={availableFilters}
+                filterGroups={filterGroups}
+                filterGroupNameMap={filterGroupNameMap}
+                groupings={groupings}
+                renderTile={renderTile}
+                hideSidebar={hideSidebar}
+              />
+              {/* <CatalogDetailsModal item={selectedItem} onClose={closeDetailsPanel} /> */}
```
rohitkrai03, line 196: Shouldn't need to comment it out.

src/shared/components/catalog/catalog-view/CatalogView.tsx
```diff
@@ -0,0 +1,269 @@
...
+
+  const filteredItems: CatalogItem[] = React.useMemo(() => {
+    const filteredByCategoryItems = filterByCategory(items, activeCategoryId, categorizedIds);
+    const filteredBySearchItems = filterBySearchKeyword(
+      filteredByCategoryItems,
+      activeSearchKeyword,
+    );
+    const filteredByAttributes = filterByAttributes(filteredBySearchItems, activeFilters);
+
+    // const filterCounts = getFilterGroupCounts(filteredBySearchItems, activeFilters, filterGroups);
```
rohitkrai03, line 173: Add a comment why this is commented out.

src/shared/components/markdown-view/markdown-view.tsx
```diff
@@ -0,0 +1,251 @@
+import * as React from 'react';
+import cx from 'classnames';
+import * as _ from 'lodash-es';
+import { Converter } from 'showdown';
+import * as sanitizeHtml from 'sanitize-html';
+import { useTranslation } from 'react-i18next';
+import { useForceRender, useResizeObserver } from '../../hooks';
+
+import './_markdown-view.scss';
```
rohitkrai03, line 9: Rename files to PascalCase.
```suggestion
import './MarkdownView.scss';
```

src/shared/style/ancillary/_bootstrap-grid.scss
```diff
@@ -0,0 +1,1002 @@
+.container {
```
rohitkrai03, line 1: Don't think we need all of these bootstrap styles. We should only bring styles that are needed for the components we are bringing in.

rohitkrai03 (changes requested):

rohitkrai03:
For some reason, the CI seems to be stuck. @divyanshiGupta Can you try force pushing again?

rohitkrai03:
Seems like the CI was getting stuck generating and sending code coverage reports. Maybe we can disable it for now and investigate it later. cc: @sawood14012

rohitkrai03:
Seems like there is an issue with the use of the `history` object the way we did in the console. We cannot do the same here because `InsightsChrome` doesn't expose the history object of `react-router`. This is why any operation related to manipulating `history` doesn't work as expected. This is the reason search is not working currently or selecting a catalog tile is not updating the history correctly.

I had a quick chat about this with @karelhala, we'll need a way to expose and share the history object for some of the functionalities to work correctly. For now, we can get this PR in and follow up with a fix. cc: @sahil143

src/shared/components/catalog/CatalogController.tsx
rohitkrai03: Discussed offline. See https://github.com/openshift/hac-dev/pull/10#issuecomment-967462839.

test.sh
```diff
@@ -5,10 +5,10 @@ if [ ! -d node_modules ]; then
     npm install
 fi
 
-curl -Os https://uploader.codecov.io/latest/linux/codecov
+# curl -Os https://uploader.codecov.io/latest/linux/codecov
```
rohitkrai03, line 8: Add a comment explaining why this is commented out.

rohitkrai03 (changes requested):

rohitkrai03 (approved):
Overall changes seem okay. There are a few things that might need follow up PRs - 
- Issue with the `history` object.
- Cleanup of unnecessary styles.
- Investigating the code coverage issue with CI.
- Adding more unit tests to improve code coverage.

Due to time constraints and upcoming PTO let's get this PR merged.

/lgtm
/approve

PR #17, opened 2021-12-02
Update eslint rules and add husky pre commit hook

rohitkrai03:


src/components/ComponentSamples/ComponentSamplesPage.tsx
```diff
@@ -108,7 +108,7 @@ export const ComponentSamplesPage = () => {
     if (formState.component) {
       setSelected(formState.component);
     }
-  }, []);
+  }, [formState.component]);
```
sahil143, line 111: @rohitkrai03  can we remove this dependency. We just need setSelected called once when the component is mounted
```suggestion
  }, []);
```

src/pages/SamplePage/SamplePage.tsx
```diff
@@ -1,11 +1,10 @@
...
-import { Link, withRouter } from 'react-router-dom';
-import { useDispatch } from 'react-redux';
-
 import { Button, StackItem, Stack, Title, Spinner } from '@patternfly/react-core';
+import { addNotification } from '@redhat-cloud-services/frontend-components-notifications/redux';
 import { Main } from '@redhat-cloud-services/frontend-components/Main';
 import { PageHeader, PageHeaderTitle } from '@redhat-cloud-services/frontend-components/PageHeader';
-import { addNotification } from '@redhat-cloud-services/frontend-components-notifications/redux';
+import { useDispatch } from 'react-redux';
+import { Link, withRouter } from 'react-router-dom';
```
sahil143, line 7: not sure but shouldn't these come after 'react' import?

src/components/ComponentSamples/ComponentSamplesPage.tsx
rohitkrai03: Done. Added a comment and disabled the eslint rule.

src/pages/SamplePage/SamplePage.tsx
rohitkrai03: Added extra path groups to handle this scenario. By default `@` groups are sorted before normal imports.

sahil143 (approved):
/lgtm

PR #22, opened 2021-12-16
Adds temp dynamic-plugin-sdk k8s utils

andrewballantyne:
## Fixes 
https://issues.redhat.com/browse/HAC-381

## Description

Adds in a `dynamic-plugin-sdk.ts` file into the root of `/src` (I can move it wherever you want though); this is effectively a blackbox to HAC-Dev. You can pry if you want & modify if you want, but exposed API is planned to be this structure come the NPM release in Jan. These are utilities to effectively open up your code to use k8s resources from DevSandbox.

Few major notes:
-  `dynamic-plugin-sdk.ts` is short lived... the file will be deleted in the long run and replaced with NPM; I have disabled eslint in this file
    - I recommend you treat this as an npm package and not play with the internals, but they are free to do so if you need to fix a function's internals or want to improve the hook... I'd recommend not changing the signature of anything exported for easy migration when the NPM package is available
- I've modified your webpack config to proxy to the K8s cluster, this is done on `/api/k8s` -- change this if you want, but you'll need to update the constant at the top of the dynamic-plugin-sdk file as well.
- I've been testing the code with this [test file / UI](https://gist.github.com/andrewballantyne/908a55ac45575f600cd7b69ae2febed5) -- feel free to use it as well or reference it to see how calls are made; recommend not checking it in (obv)
    - Ask questions if you see something here you're not sure about, any SDK API is intended to be a future API, so if there are issues, please speak up
- You'll run your UI through this command: `REMOTE_CONFIG=ci ENVIRONMENT=prod BETA=true PROXY=true npm run start` -- this should give you access to the Prod SSO token that is used for AppStudio DevSandbox, which is pulled and used inside the `dynamic-plugin-sdk.ts` file / API calls
    - Be sure you [sign up properly](https://coreos.slack.com/archives/C02CTEB3MMF/p1639510979289400?thread_ts=1639475390.241600&cid=C02CTEB3MMF) to get your SSO token tied to the DevSandbox instance otherwise it will deny you

**The SDK file is big and messy... What exactly is in `dynamic-plugin-sdk.ts` for me to use?**

1. K8s utility calls
    1. `k8sGetResource`
    2. `k8sCreateResource`
    3. `k8sUpdateResource`
    4. `k8sPatchResource`
    5. `k8sDeleteResource`
    6. `k8sListResource`
2. A hook to watch a resource, `useK8sWatchResource`
    - It should be noted this is a hacked version; the real one uses apiDiscovery data that we do not have ready nor is it part of this offering
    - It assumes weird pluralization because of this gap (Model's `plural: initResource.groupVersionKind.kind.toLowerCase() + 's',`) -- this is far from ideal, but hopefully this covers your use-cases for the short term while we work through the rest of the dependencies in the SDK
3. Types
    1. `K8sResourceCommon` - basic as it gets, no other K8s resource defs are provided
    2. `K8sModel` - all the APIs work off of this, we'll need to work on how this is offered as an SDK alternative, but this for now is a type you can do if you want to craft your own models
    3. `K8sGroupVersionKind` - this is the new `core~pods~v1` approach the SDK is taking on... it's an object of the 3 props internally
    4. The two types for the hook: `WatchK8sResource` & `WatchK8sResult<R>`

TODO:
- [x] Fix up the hook so it performs as it should -- can get into infinite update states and hook firing

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

No screenshots, it's configuration & a utility file. But the UI I used to test is:

image
Upon calling list, it tries to [determine your namespace](https://gist.github.com/andrewballantyne/908a55ac45575f600cd7b69ae2febed5#file-testk8s-tsx-L91) based on what you have access to. This will help with the rest of the calls (hence them being enabled). 
image

## How to test or reproduce?

N / A... not bringing any new UI.

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/dynamic-plugin-sdk.ts
+++ b/src/dynamic-plugin-sdk.ts
@@ -1,0 +1,554 @@
+/* eslint-disable */
+import * as React from 'react';
+import * as _ from 'lodash';
+
+/* throw away fetch methods */
+const k8sBasePath = `/api/k8s`; // webpack proxy path
+
+const commonFetch = async (url, method, options, timeout): Promise<any> => {
+  // Grab the token out of the cookie for ConsoleDot | if not proxying to prod, token will be invalid even if it is there
+  const cookieToken = document.cookie.split('; ').find((val) => val.startsWith('cs_jwt='));
+  if (!cookieToken) {
+    return Promise.reject('Could not make k8s call. Unable to find token.');
+  }
+  const [,token] = cookieToken.split('=');
+
+  const allOptions = _.defaultsDeep({ method }, options, { headers: { Accept: 'application/json' } });
+  allOptions.headers.Authorization = `Bearer ${token}`;
+  try {
+    const response = await fetch(url, allOptions);
+    if (response.status === 401) {
+      return Promise.reject('Invalid token. Are you working with Prod SSO token?');
+    }
+    const json = await response.json();
+    return Promise.resolve(json);
+  } catch(e) {
+    return Promise.reject(e);
+  }
+};
+
+const consoleFetchSendJSON = (
+  url: string,
+  method: string,
+  json = null,
+  options: RequestInit = {},
+  timeout: number,
+) => {
+  const allOptions: Record<string, any> = {
+    headers: {
+      Accept: 'application/json',
+      'Content-Type': `application/${
+        method === 'PATCH' ? 'json-patch+json' : 'json'
+      };charset=UTF-8`,
+    },
+  };
+  if (json) {
+    allOptions.body = JSON.stringify(json);
+  }
+  return commonFetch(url, method, _.defaultsDeep(allOptions, options), timeout);
+};
+
+const coFetchJSON = (...args) => commonFetch.apply(undefined, args);
+coFetchJSON.post = (url, json) => consoleFetchSendJSON(url, 'POST', json, undefined, undefined);
+coFetchJSON.put = (url, json) => consoleFetchSendJSON(url, 'PUT', json, undefined, undefined);
+coFetchJSON.patch = (url, json) => consoleFetchSendJSON(url, 'PATCH', json, undefined, undefined);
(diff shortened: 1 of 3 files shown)


src/dynamic-plugin-sdk.ts
```diff
@@ -0,0 +1,554 @@
+/* eslint-disable */
+import * as React from 'react';
+import * as _ from 'lodash';
+
```
rohitkrai03, line 4: @sawood14012 Hey, can we not have this codecov thing popup on every line in the PR? Its really distracting while doing the review.

sawood14012: you mean the codcov bot ? i guess it is added by openshift org you can configure it in the repo settings telling it to only post once .

christianvogt: @rohitkrai03 you can hide them using the `Show annotations` action:
image

andrewballantyne:
Removing WIP header as it seems to be deterring engagement on this PR. There is still one last thing to do, but it shouldn't stop collaboration on the main purpose of this PR, which is that HAC-Dev can get access to DevSandbox via the work here.

andrewballantyne:
Updated the testFile ([external example file](https://gist.github.com/andrewballantyne/908a55ac45575f600cd7b69ae2febed5)) and the hook code (part of this PR).

src/dynamic-plugin-sdk.ts
```diff
@@ -0,0 +1,603 @@
...
+const HOOK_POLL_DELAY = 2000; // change this if you want the useHook to go faster / slower on polls
```
andrewballantyne, line 5: If for whatever reason 2 seconds is too long or too short, you can just modify this constant to change the poll rate of the hook.

rohitkrai03 (approved):
/lgtm
/approve

PR #25, opened 2021-12-27
Add support for importing public git repository into application

rottencandy:
## Fixes 
https://issues.redhat.com/browse/HAC-370

## Description
Adds the ability to create components using a public git repository.

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


src/components/AddComponent/AddComponentForm.tsx
```diff
@@ -5,10 +5,30 @@ import isEmpty from 'lodash/isEmpty';
...
+import { GitOptions } from './GitOptions';
 import { SourceField } from './SourceField';
 
 export type AddComponentValues = {
   source: string;
+  git: {
+    reference: string;
+    contextDir: string;
+  };
+  detectedComponents?: {
```
rohitkrai03, line 17: Is this based on the latest kube API specs for `ComponentDetectionQuery`? I think there is a component resource stub that has all the default values for component resource. Ref - https://docs.google.com/document/d/1CBeRA8LrdsDBG4AqFYqn4aovcd969lP_1-41OUAN2Bk/edit#heading=h.apl9rlcsyxg8

src/shared/components/name-value-editor/_name-value-editor.scss
```diff
@@ -0,0 +1,72 @@
+$drag-column-width: 28px;
```
rohitkrai03, line 1: Don't name files starting with `_`. Should be named as the component its being used in.

src/components/AddComponent/AddComponentForm.tsx
rottencandy: Yes I had mocked the ComponentDetectionQuery data from that doc in `create-utils.ts`: https://github.com/openshift/hac-dev/pull/25/files#diff-18923f7bff33b78fd8fb3983b01068eb30ac83937137a1580f332ee7560e3361R87-R128

src/shared/components/name-value-editor/_name-value-editor.scss
rottencandy: I copied it from ODC which had the `_`, removed it.

rohitkrai03 (changes requested):
@rottencandy On first run I found a few issues -
- I think we need to run basic validation in the UI for valid URL even before we validate the git repo using the HAS API. Right now it's taking any random input as valid input.
- The source URL that user is entering is not being used in the review component form.
- The component name should be derived from the source URL itself.
- The edit button on the component name doesn't work.
- I don't think the Component CR supports health checks as of now so e can skip that in the review form for now. Refer - https://coreos.slack.com/archives/C02CTEB3MMF/p1638977088198300 and https://docs.google.com/document/d/1dfkTky3qFv6wTXdMX9Ag-bYFEt3OfBgRZOceYNSwyEU/edit#heading=h.slc3w33b9eji. 
- Clicking on the whole `Build Configuration` or `Deploy Configuration` doesn't expand the section. It only works on the toggle icon.

rottencandy:
@rohitkrai03 have a couple of questions
> * The source URL that user is entering is not being used in the review component form.

I was showing the url provided by the component detection query's data (in `spec.source.git.url`) which might be different for every component. Should we be showing the source URL in the card title instead?
Or did you mean that we need to show the URL in a separate field?


> * The component name should be derived from the source URL itself.

Same here, wouldn't the name be the same for all components? I thought we had to use the name provided in the CDQ data.

rohitkrai03:
> @rohitkrai03 have a couple of questions
> 
> > * The source URL that user is entering is not being used in the review component form.
> 
> I was showing the url provided by the component detection query's data (in `spec.source.git.url`) which might be different for every component. Should we be showing the source URL in the card title instead? Or did you mean that we need to show the URL in a separate field?
> 
> > * The component name should be derived from the source URL itself.
> 
> Same here, wouldn't the name be the same for all components then? I thought we had to use the name provided in the CDQ data.

Okay, you might be right there. it makes sense to use data returned from detection query in this case even though in most of the case it might be same as git source url.

rottencandy:
@rohitkrai03 addressed all the issues.

src/models/component.ts
```diff
@@ -10,3 +10,11 @@ export const ComponentModel: K8sModel = {
...
   namespaced: true,
 };
+
+export const HASComponentDetectionQueryModel = {
+  group: 'has.appstudio',
+  version: 'v1alpha1',
+  kind: 'ComponentDetectionQuery',
+  abbr: 'CDQ',
+  singular: 'componentdetectionquery',
+};
```
rohitkrai03, line 20:
```suggestion
export const ComponentDetectionQueryModel = {
  group: 'appstudio.redhat.com',
  version: 'v1alpha1',
  kind: 'ComponentDetectionQuery',
  abbr: 'CDQ',
  plural: 'componentdetectionqueries',
  label: 'Component Detection Query',
  labelPlural: 'Component Detection Queries',
  namespaced: true,
};
```

src/shared/style.scss
```diff
@@ -4,6 +4,11 @@
 // Base
 @import 'style/common';
 @import 'style/layout';
+@import 'style/icons';
+
+// grid
+@import 'style/bootstrap-grid';
```
rohitkrai03, line 10: Why do we need bootstrap grid?

rohitkrai03 (changes requested):
Some issues I found in the styling - 

- The review components form shows  grey background in the marvel doc UX provided.
- The expandable sections do not have any indentation which is there in the marvel doc.
- The component name seems a bit off and not inline with the toggle. You might need to remove the line-height from there and add margin-bottom instead.
- Some of the sections in review form seems a bit cramped. Might need some margins and paddings.

src/shared/style.scss
rottencandy: The `col-*` classes are used in the name value editor component.

src/components/AddComponent/GitOptions.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import { ExpandableSection, Stack, StackItem } from '@patternfly/react-core';
+import { useField } from 'formik';
+import { InputField } from '../../shared';
+
+export const GitOptions: React.FC = () => {
+  const [, { value: detectedComponents }] = useField('detectedComponents');
+  if (!detectedComponents) {
```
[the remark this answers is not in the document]
rottencandy: I was a bit confused because in the marvel design, the section appears after the URL is entered. So I thought It had to be shown after validation.
Updated.

src/components/AddComponent/SourceField.tsx
```diff
@@ -5,34 +5,92 @@ import {
...
+export const SourceField: React.FC<SourceFieldProps> = ({ onSamplesClick }) => {
+  const [, { value: source, touched, error }] = useField<string>({
+    name: 'source',
+    type: 'input',
+  });
+  const [, , { setValue: setDetectedComponents }] = useField('detectedComponents');
+  const [validated, setValidated] = React.useState(ValidatedOptions.default);
+  const [helpText, setHelpText] = React.useState('');
+  const [formState] = useFormValues();
+  useFormikValidationFix(source);
```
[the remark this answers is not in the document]
rottencandy: Yeah, validation issue doesn't seem to happen anymore, I had not tested properly. Thanks.

rottencandy:
@rohitkrai03 updated. PTAL.

src/components/AddComponent/GitOptions.tsx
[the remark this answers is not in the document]
rohitkrai03: It needs to be shown after validation of the URL in UI. Not validated by the detection query.

rohitkrai03 (approved):
/lgtm
/approve

Misjohns:
@rottencandy This looks good, just caught a few minor tweaks that we'd like to get updated for this sprint since it's already merged. Please let me know if you'd like a quick call to discuss. Thank you.

Build your application:
1. Change the "Component source" field label to "Git repo URL or container image"
2. Add a period to the end of "Start with a sample." 
3. Change the "Show options" section title to "Authorization and Git options”

Review your new components:
1. Change “choices” to “selections” so the description reads, “Review your selections for the application.”
2. The Runtime dropdown should show the icon for the software and the text should include the name and version (ie. Node.js 10.1)
3. There needs to be more space between the Run command field and the Deploy configuration section header. Roughly 32px.
4. For CPU and Memory please make the “C” in “container” lowercase and correct spelling of “Guaranteed”.

Beau is going to create a Jira task to capture non-happy path items such as Invalid URLs, taken names, etc.

PR #26, opened 2022-01-07
Add HAS API integration to samples flow

rohitkrai03:
## Fixes 

https://issues.redhat.com/browse/HAC-405

## Description

This PR change removes mock data usage from samples flow and integrates it with HAS API deployed on devsanbox proxy.

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

https://user-images.githubusercontent.com/6041994/148555929-e12db0e2-2c1c-43a7-8654-225efc57616a.mov

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


src/yamls/app.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: appstudio.redhat.com/v1alpha1
```
rottencandy, line 1: Where would these yamls be used?

src/pages/TestK8s.tsx
```diff
@@ -0,0 +1,231 @@
...
+        setStatus(`${action} response:`);
+        setR(data);
+        console.debug(`++++${action}!`, data);
+      })
+      .catch((err) => {
+        console.error(`++++failed ${action}`, err);
+        setStatus('failed call');
+        setR(null);
+      });
+  }, [action, name, namespace]);
```
rottencandy, line 204: Not a big deal since this is just a test page, but this effect is firing on every keystroke. Is it okay for it to be that way? Maybe we can't do much about it

rottencandy (approved):
Tested locally, works as expected. Code changes also look good.
Just had a couple of questions.

src/yamls/app.yaml
rohitkrai03: Its just a sample to understand the APIs right now. We can remove them later.

src/pages/TestK8s.tsx
rohitkrai03: Yeah, we don't need to worry about this right now. Its just for testing APIs.

PR #43, opened 2022-02-15
Add authorization for private git repositories

rottencandy:
## Fixes 
https://issues.redhat.com/browse/HAC-647

## Description
Add git repo authorization using `SPIAccessTokenBinding` and use provided secret to run component detection.

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

https://user-images.githubusercontent.com/20013884/155485712-b1e3576d-8729-4c7f-bf71-45720050e651.mp4

## How to test or reproduce?
Use a private repository, and select authorization in git options.
(Note that CDQ does not support private repositories at this point, so detection will always fail)

## Browser conformance: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/AddComponent/GitAuthorization.tsx
+++ b/src/components/AddComponent/GitAuthorization.tsx
@@ -1,0 +1,71 @@
+import * as React from 'react';
+import {
+  Button,
+  ButtonVariant,
+  HelperText,
+  HelperTextItem,
+  Popover,
+  Text,
+  TextContent,
+  TextVariants,
+} from '@patternfly/react-core';
+import { CheckIcon } from '@patternfly/react-icons/dist/js/icons/check-icon';
+import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons/dist/js/icons/outlined-question-circle-icon';
+import { useField } from 'formik';
+import { useActiveNamespace } from '../../hooks/useActiveNamespace';
+import { LoadingInline } from '../../shared/components/status-box/StatusBox';
+import { initiateAccessTokenBinding } from '../../utils/create-utils';
+import { useAccessTokenBindingAuth } from './utils';
+
+export const GitAuthorization: React.FC = () => {
+  const namespace = useActiveNamespace();
+  const [isCreating, setIsCreating] = React.useState(false);
+  const [bindingName, setBindingName] = React.useState('');
+  const [, { value: url, error, touched }] = useField<string>('source');
+  const [, { value: authSecret }] = useField<string>('git.authSecret');
+
+  useAccessTokenBindingAuth(bindingName);
+
+  const startAuthorization = React.useCallback(() => {
+    setIsCreating(true);
+    initiateAccessTokenBinding(url, namespace)
+      .then((resource) => {
+        setBindingName(resource.metadata.name);
+      })
+      .finally(() => {
+        setIsCreating(false);
+      });
+  }, [namespace, url]);
+
+  return (
+    <TextContent>
+      <Text component={TextVariants.h3}>
+        Authorize Git repository{' '}
+        <Popover bodyContent="Authorizing private Git repositories allows App Studio to access your source code.">
+          <OutlinedQuestionCircleIcon />
+        </Popover>
+      </Text>
+      <Text component={TextVariants.small}>
+        To connect to a private repository, authorize access to pull in source code.
+      </Text>
+      {authSecret ? (
+        <HelperText>
+          <HelperTextItem variant="success" icon={<CheckIcon />}>
+            Authorized access
+          </HelperTextItem>
+        </HelperText>
+      ) : isCreating ? (
+        <LoadingInline />
+      ) : (
+        <Button
+          variant={ButtonVariant.link}
+          onClick={startAuthorization}
+          isDisabled={!touched || !!error}
+          style={{ paddingLeft: 0 }}
+        >
+          Sign In
+        </Button>
+      )}
+    </TextContent>
+  );
+};
--- a/src/utils/create-utils.ts
+++ b/src/utils/create-utils.ts
@@ -1,7 +1,13 @@
+import isEqual from 'lodash/isEqual';
 import uniqueId from 'lodash/uniqueId';
-import { ApplicationModel, ComponentModel, ComponentDetectionQueryModel } from '../models';
-import { ApplicationKind, ComponentDetectionQueryKind } from '../types';
-import { k8sCreateResource, k8sGetResource } from './../dynamic-plugin-sdk';
+import { k8sCreateResource, k8sGetResource, k8sListResource } from '../dynamic-plugin-sdk';
+import {
+  ApplicationModel,
+  ComponentModel,
+  ComponentDetectionQueryModel,
+  SPIAccessTokenBindingModel,
+} from '../models';
+import { ApplicationKind, ComponentDetectionQueryKind, SPIAccessTokenBindingKind } from '../types';
 
 /**
  * Create HAS Application CR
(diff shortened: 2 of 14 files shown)


rottencandy:
> @rottencandy Shouldn't this open up the oauth window and go through that flow to get authorised? And I think we need to show some kind of success notification or something in the form.

It does open a new tab for oauth, but looks like it is a one-time process and I couldn't reset the authorization to run it again, so it directly provides the secret if authentication has been done previously.

Added notification:

https://user-images.githubusercontent.com/20013884/154218119-b47df204-a0c2-4702-8f65-0aaca72b14aa.mp4

src/utils/create-utils.ts
```diff
@@ -151,3 +158,53 @@ export const createComponentDetectionQuery = async (
...
+  };
+
+  let binding: SPIAccessTokenBindingKind = await k8sCreateResource({
+    model: SPIAccessTokenBindingModel,
+    data: requestData,
+  });
+
+  while (!binding.status) {
+    await new Promise((r) => setTimeout(r, 200));
+    binding = await k8sGetResource({
```
christianvogt, line 202: What's the point polling here? Seems odd when we should be using a watch.

rottencandy: Updated to use watch.

src/utils/create-utils.ts
```diff
@@ -151,3 +158,42 @@ export const createComponentDetectionQuery = async (
...
+    spec: {
+      repoUrl: url,
+      permissions: {
+        required: [
+          { type: 'r', area: 'repository' },
+          { type: 'w', area: 'repository' },
+        ],
+      },
+      secret: {
+        name: 'hac-dev-repo-token',
```
christianvogt, line 187: Won't this secret end up clashing with other components that require auth?

src/components/AddComponent/GitAuthorization.tsx
```diff
@@ -0,0 +1,59 @@
...
+  TextVariants,
+} from '@patternfly/react-core';
+import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons/dist/js/icons/outlined-question-circle-icon';
+import { useField } from 'formik';
+import { useActiveNamespace } from '../../hooks/useActiveNamespace';
+import { LoadingInline } from '../../shared/components/status-box/StatusBox';
+import { initiateAccessTokenBinding } from '../../utils/create-utils';
+import { useAccessTokenBindingAuth } from './utils';
+
+export const GitAuthorization: React.FC = () => {
```
christianvogt, line 17: If someone initiates oauth flow, but then closes the window. How do they re-establish the auth flow with the existing request binding?
It doesn't seem like that happens now (although I haven't tried this PR).
Also what about cleaning up `SPIAccessTokenBinding` that are no longer relevant? for example if the user changes the git repo URL after going through the auth flow, or leaving the page without creating the component.

src/utils/create-utils.ts
rottencandy: Appended uid to secret name.

src/components/AddComponent/GitAuthorization.tsx
rottencandy: Right now the url and access binding context is only stored in formik state, and we'd lose all of it if the window is closed.
Should we save this context in local storage maybe?

Also note that users don't have edit/delete access to `SPIAccessTokenBinding`s, so not sure how we'd clean it up after creation.

rohitkrai03:
@rottencandy Take a look at the final designs for this flow. They've changed a bit.

src/components/AddComponent/GitAuthorization.tsx
rottencandy: Updated.

The hook will now look for an existing `SPIAccessTokenBinding` for the ULR before creating a new one and use that instead if available.
And the created `SPIAccessTokenBinding` will be deleted on unmount.

rottencandy:
> @rottencandy Take a look at the final designs for this flow. They've changed a bit.

Updated to align with designs.
cc @misjohns @beaumorley

src/utils/create-utils.ts
```diff
@@ -151,3 +160,73 @@ export const createComponentDetectionQuery = async (
...
+ * @returns Returns the SPIAccessTokenBinding resource
+ */
+export const initiateAccessTokenBinding = async (url: string, namespace: string) => {
+  const bindings: SPIAccessTokenBindingKind[] = await k8sListResource({
+    model: SPIAccessTokenBindingModel,
+    queryParams: {
+      ns: namespace,
+      labelSelector: {
+        matchLabels: {
+          [SPI_AUTH_REPO_LABEL]: 'hac-dev',
```
christianvogt, line 221: Is this label match necessary? What if a binding was created from CLI instead?

src/utils/create-utils.ts
```diff
@@ -151,3 +160,73 @@ export const createComponentDetectionQuery = async (
...
+        },
+      },
+    },
+  });
+  const binding = bindings.find((b) => b.spec.repoUrl === url);
```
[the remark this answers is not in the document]
christianvogt: reusing the binding is preferred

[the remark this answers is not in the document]
rottencandy: Added check for comparing permissions.

src/utils/create-utils.ts
rottencandy: Didn't realize about it before, I think the label doesn't add any value, removed it.

src/utils/create-utils.ts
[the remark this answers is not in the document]
rottencandy: @christianvogt since it's preferred to reuse bindings, should we not delete any binding when moving away from the page? Or maybe delete only the ones created by UI?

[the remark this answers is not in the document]
christianvogt: That's a good question.
I hate to leave around resources when the flow wasn't completed however it probably doesn't hurt since we will be doing a lookup to reuse.
Let's leave them for now and we can revisit this later.

src/components/AddComponent/GitAuthorization.tsx
christianvogt: Deleting on unmount would never have worked because unmount could occur whether they completed the form or not. The idea was only to delete if they left the flow without completing it.

As per another comment, since we are reusing existing bindings, let's keep them around for now and revisit this later.

src/utils/create-utils.ts
```diff
@@ -151,3 +159,75 @@ export const createComponentDetectionQuery = async (
...
+ * @param namespace namespace to create the binding
+ * @returns Returns created SPIAccessTokenBinding resource
+ */
+export const createAccessTokenBinding = async (url: string, namespace: string) => {
+  const id = uid();
+  const requestData = {
+    apiVersion: `${SPIAccessTokenBindingModel.apiGroup}/${SPIAccessTokenBindingModel.apiVersion}`,
+    kind: SPIAccessTokenBindingModel.kind,
+    metadata: {
+      name: `hac-dev-import-${id}`,
```
christianvogt, line 176: We don't want to introduce `hac-dev` as a prefix. 
Maybe use `appstudio` instead unless there's another name given to appstudio that's more user facing?

src/utils/create-utils.ts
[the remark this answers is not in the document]
rottencandy: Removed deletion hook.

src/utils/create-utils.ts
rottencandy: Updated to use the `appstudio` prefix.

christianvogt:
/approve

Please have someone test the PR for `lgtm`.

As a followup, it would be good to automatically attempt authorization if we find a matching token binding since we are reusing them. Instead of waiting for the user to expand the section and click the button. cc @beaumorley

rohitkrai03 (approved):
/lgtm

Tested this locally with multiple private repos. HAS API doesn't support private repos yet so detection still fails but that support should come soon from HAS.

The oauth flow gets access to all private repos of a user and not just the entered GIt Repo URL. This means we can use the same binding for a user with different private repos. Good thing is SPI doesn't go through whole oauth flow for new private repo URL of same user. For now I think its okay to reuse the same binding per URL but maybe we can improve this later with same binding per user.

PR #44, opened 2022-02-15
Add unit tests for catalog components and utils

vikram-raj:
## Fixes 
https://issues.redhat.com/browse/HAC-398

## Description
Add unit tests for catalog components and utils

## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Build related changes
- [x] Other (please describe): adds unit tests


src/shared/components/catalog/__tests__/CatalogTile.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+    label: 'Sample',
+    value: 'Sample',
+    description: '',
+  },
+];
+
+describe('CatalogTile', () => {
+  it('should show tag badges on catalog tile', () => {
+    render(<CatalogTile item={catalogItems[0]} />);
+    expect(screen.queryByTestId('tag-badges')).not.toBeNull();
```
invincibleJai, line 24: as we are using `queryByTestId` so can we check to `toHaveLength`?

src/shared/components/catalog/__tests__/CatalogView.spec.tsx
```diff
@@ -0,0 +1,50 @@
...
+  });
+
+  it('should not render Catalog filters', () => {
+    render(<CatalogView items={catalogItems} renderTile={renderTile} />);
+    expect(screen.queryByTestId('catalog-filters')).toBeNull();
+  });
+
+  it('should not render Catalog grid', () => {
+    render(<CatalogView items={[]} renderTile={renderTile} />);
+    expect(screen.queryByTestId('catalog-grid')).toBeNull();
```
invincibleJai, line 47: we would need `queryByTestId` or `getByTestId`?

vikram-raj: I think using `queryByTestId` is fine as 'catalog-grid' is not going to be present in the DOM. If I use getByTestId` it will throw an error.

rottencandy (approved):
/lgtm
Using `toBeInTheDocument()` would have been cleaner than `.toBeNull()` but it's a minor concern and current changes are fine too.

PR #45, opened 2022-02-15
Add delete action in the Component list page

debsmita1:
## Fixes 
https://issues.redhat.com/browse/HAC-544

## Description
- Added Delete action in the kebab menu which on clicked will delete the component CR
- Added tests

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

delete-components

## How to test or reproduce?
Goto: https://prod.foo.redhat.com:1337/beta/hac/app-studio/components?application=<application-name>

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/ComponentListView/ComponentListViewPage.tsx
+++ b/src/components/ComponentListView/ComponentListViewPage.tsx
@@ -1,16 +1,14 @@
 import * as React from 'react';
 import { Link } from 'react-router-dom';
 import {
+  Button,
   DataList,
-  DataListCell,
   DataListItem,
-  DataListItemCells,
   DataListItemRow,
   Toolbar,
   ToolbarContent,
   ToolbarItem,
   TextInput,
-  Button,
 } from '@patternfly/react-core';
 import { FilterIcon } from '@patternfly/react-icons/dist/js/icons';
 import { useK8sWatchResource } from '../../dynamic-plugin-sdk';
@@ -31,6 +29,7 @@ export const ComponentListViewPage: React.FC<ComponentListViewPageProps> = ({ ap
     namespace,
     isList: true,
   });
+  const [nameFilter, setNameFilter] = React.useState<string>('');
   const loaded = namespace && componentsLoaded;
 
   // TODO: handle empty state / components not found
@@ -38,8 +37,6 @@ export const ComponentListViewPage: React.FC<ComponentListViewPageProps> = ({ ap
     () => (loaded ? allComponents?.filter((c) => c.spec.application === application) : []),
     [allComponents, application, loaded],
   );
-
-  const [nameFilter, setNameFilter] = React.useState<string>('');
 
   const filteredComponents = React.useMemo(
     () =>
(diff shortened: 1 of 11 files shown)


src/components/ComponentListView/ComponentListViewPage.tsx
```diff
@@ -51,48 +53,95 @@ export const ComponentListViewPage: React.FC<ComponentListViewPageProps> = ({ ap
...
+      : updatedCheckedColumns.add(selectedItem);
+    setSelectedComponents(updatedCheckedColumns);
+  };
+
+  const deleteSelectedRows = async () => {
+    await Promise.all(
+      Array.from(selectedComponents).map((component) =>
+        deleteComponent(component, application, namespace),
+      ),
+    )
```
rottencandy, line 73: These callbacks are making use of component state, would be good to memoize them.

src/utils/delete-utils.ts
```diff
@@ -0,0 +1,34 @@
...
+    },
+    spec: {
+      application,
+    },
+  };
+  // TODO: Make Api Calls here
+  return k8sDeleteResource({
+    model: ComponentModel,
+    resource: componentData,
+  });
```
rottencandy, line 33: Not an issue but seems strange that we have to pass the component's entire spec just to delete it :thinking:

src/components/ComponentListView/ComponentListViewPage.scss
```diff
@@ -0,0 +1,7 @@
+.hacDev-component-list-view-page__trash-icon {
+  color: var(--pf-global--BackgroundColor--dark-400);
+  &:hover {
+    cursor: pointer;
```
rottencandy, line 4: `<Button>` doesn't make it pointer by default?

src/utils/delete-utils.ts
debsmita1: removed the spec

src/components/ComponentListView/ComponentListViewPage.scss
debsmita1: Thanks for catching this. Removed the css

src/utils/delete-utils.ts
rottencandy: @debsmita1 we won't need the `application` arg now.

rottencandy (commented):
When pressing delete, there is no indication that components are being deleted, maybe we can use `isLoading` for the button to show a spinner while the deletion happens. WDYT?

Also, not sure if it would be good should show a notification once deletion is completed.

https://user-images.githubusercontent.com/20013884/154287585-ae7be36b-d02d-4259-8f82-c6a3d2ce1375.mp4

debsmita1:
> When pressing delete, there is no indication that components are being deleted, maybe we can use `isLoading` for the button to show a spinner while the deletion happens. WDYT?
> 
> Also, not sure if it would be good should show a notification once deletion is completed.
> 
>  0.mp4

^ @beaumorley @rohitkrai03

rohitkrai03:
> > When pressing delete, there is no indication that components are being deleted, maybe we can use `isLoading` for the button to show a spinner while the deletion happens. WDYT?
> > Also, not sure if it would be good should show a notification once deletion is completed.
> > 0.mp4
> 
> ^ @beaumorley @rohitkrai03

@debsmita1 Can you start a conversation around this in slack to get faster turnaround?

debsmita1:
> @debsmita1 You might have to rework this a bit because designs have changes around this page.

@rohitkrai03  Updated. PTAL !

rottencandy (approved):
/lgtm

rohitkrai03 (approved):
@debsmita1 Overall looks good but I think we should also add the delete confirmation modal similar to what we have in console. Its not specified in the UX designs but I think its good pattern to follow here. I think we'll need the modal launcher from console but its not in a good state to bring in as is. Maybe we do this as a follow up story.

PR #47, opened 2022-03-03
Extension port

karelhala:
## Description
This will make hac-dev as plugin in hac-core. There is no longer AppEntry and we do not wrap components in providers since all of that should be done in hac-core.

## Type of change

- [x] Feature

## How to test or reproduce?
It should be enough just to install latest deps and run `ENVIRONMENT=prod npm run start:beta` (we have one last PR to fully support this https://github.com/openshift/hac-core/pull/35)

## Known limitations
There are currently two know limitations:
- no dry run for create resource (@vojtechszocs  @andrewballantyne are there any plans in supporting this?)
- no redux, we are working on adding redux to hac-core and will keep you updated on the process


karelhala:
@andrewballantyne oooh, will do that!

config/dev.webpack.config.js
```diff
@@ -64,7 +57,7 @@ const webpackProxy = {
 
 const { config: webpackConfig, plugins } = config({
   rootFolder: resolve(__dirname, '../'),
-  sassPrefix: '.hacDev',
+  sassPrefix: '.hacCore',
```
andrewballantyne, line 60: I don't think this change is intended...

config/prod.webpack.config.js
```diff
@@ -5,7 +5,8 @@ const commonPlugins = require('./plugins');
 
 const { config: webpackConfig, plugins } = config({
   rootFolder: resolve(__dirname, '../'),
-  ...(process.env.BETA && { deployment: 'beta/apps' }),
+  sassPrefix: '.hacCore',
```
andrewballantyne, line 8: Think we want the other prefix here too.

package.json
```diff
@@ -15,25 +15,25 @@
     "lint:ts": "eslint . --ext .js,.jsx,.ts,.tsx --color",
     "lint:ts:fix": "npm run lint:ts -- --fix",
     "lint:sass": "stylelint 'src/**/*.scss' --config .stylelintrc.json",
-    "nightly": "npm run deploy",
```
andrewballantyne, line 18: Don't think we want to remove scripts they added @karelhala ...

src/Routes.tsx
```diff
@@ -1,60 +0,0 @@
...
-      </Bullseye>
-    }
-  >
-    <Switch>
-      <Route path="/" component={AppFlow} exact />
-      <Route path="/create" component={SamplesFlow} exact />
-      <Route path="/components" component={ComponentListView} exact />
-      <Route path="/applications" component={ApplicationList} exact />
-      <Route path="/sample-page" component={SamplePage} exact />
-      <Route path="/k8s-util" component={K8sPage} exact />
```
andrewballantyne, line 53: I _think_ you're missing `/create` and `/applicatons` routes in the extension definition?

package.json
```diff
@@ -45,18 +45,15 @@
     "react-dom": "17.0.2",
     "react-helmet": "^6.1.0",
     "react-i18next": "^11.12.0",
-    "react-redux": "7.2.4",
     "react-router-dom": "5.2.0",
     "react-virtualized": "^9.22.3",
-    "redux": "4.1.0",
-    "redux-logger": "3.0.6",
-    "redux-promise-middleware": "6.1.2",
```
andrewballantyne, line 53: @karelhala This seems a little aggressive to modify their redux usage. It's fair to remove it because they can't really use Redux because they don't own the context. But I suspect you just pulled a feature out from under them 🤔 

@rohitkrai03 what's the use-case for your redux usage? Can you guys pivot to a React Context? That's a supported extension type and should persist between routes.

src/hooks/useActiveNamespace.ts
```diff
@@ -9,7 +9,7 @@ export const useActiveNamespace = () => {
 
   useEffect(() => {
     const fetchProjects = async () => {
-      const projects = await k8sListResource({
+      const { items: projects }: any = await k8sListResource({
```
andrewballantyne, line 12: There is `k8sListResourceItems`: https://github.com/openshift/dynamic-plugin-sdk/blob/main/packages/lib-utils/src/k8s/k8s-resource.ts#L152

No need to destruct it.

src/hooks/useApplicationsInfo.ts
```diff
@@ -18,7 +18,10 @@ export const useApplicationsInfo = (): ApplicationInfo => {
   React.useEffect(() => {
     let mounted = true;
     const fetchApplication = async () => {
-      const appData = await k8sGetResource({ model: ApplicationModel, ns: namespace });
+      const appData: any = await k8sGetResource({
```
andrewballantyne, line 21: @rohitkrai03 Do you guys have K8sResourceCommon-esk types?

src/pages/TestK8s.tsx
```diff
@@ -69,6 +68,11 @@ const TestK8s: React.FC = () => {
         namespace,
       },
     };
+
+    const testConfigMetadata = {
```
andrewballantyne, line 72: ConfigMaps don't exist anymore. We should bring over our Application changes if they want to keep this page.

src/utils/__tests__/create-utils.spec.ts
```diff
@@ -11,7 +11,7 @@ import {
 
 jest.mock('../../dynamic-plugin-sdk');
```
andrewballantyne, line 12: Still mocking the old file...

src/utils/create-utils.ts
```diff
@@ -209,9 +236,9 @@ export const createAccessTokenBinding = async (url: string, namespace: string) =
  * @returns Returns the SPIAccessTokenBinding resource
  */
 export const initiateAccessTokenBinding = async (url: string, namespace: string) => {
-  const bindings: SPIAccessTokenBindingKind[] = await k8sListResource({
+  const { items: bindings }: { items: SPIAccessTokenBindingKind[] } = await k8sListResource({
```
andrewballantyne, line 239: Same point here about `k8sListResourceItems`

src/utils/delete-utils.ts
```diff
@@ -11,17 +11,23 @@ import { k8sDeleteResource } from './../dynamic-plugin-sdk';
...
-    kind: 'Component',
-    metadata: {
+  // const componentData = {
+  //   apiVersion: `${ComponentModel.apiGroup}/${ComponentModel.apiVersion}`,
+  //   kind: 'Component',
+  //   metadata: {
+  //     name: componentName,
+  //     namespace,
+  //   },
+  // };
```
andrewballantyne, line 21: Commented code?

andrewballantyne (changes requested):
Top level review. Not done to any real detail or via testing the code.

config/dev.webpack.config.js
karelhala: Well the thing is that console.dot is wrapping each app with its prefix. Since this is not directly an app, but extension it inherits the prefix of `hacCore` but I guess we could update it as `.hacCore, .hacDev`. Or (which might be better imo) wrapping each `route` with `Container` with the app's name in `hac-core`

package.json
karelhala: This is inherited from `starter-app` we are no longer doing these builds.

package.json
karelhala: I've looked trough the code and currently redux is used to show notification messages. Until we have full support of redux I don't think we can keep these in.

src/hooks/useActiveNamespace.ts
karelhala: Oooh! Okay, will do that!

src/utils/delete-utils.ts
karelhala: Aaah, yeah!

config/dev.webpack.config.js
karelhala: Opened a PR in hac-core to allow passing className as wrapper of current route. This way one route styling won't affect styling of different route.

src/Routes.tsx
karelhala: Ah good point!

rohitkrai03 (changes requested):

src/hooks/useApplicationsInfo.ts
andrewballantyne: @rohitkrai03 You will get common types for sure... but use-case specific types is beyond the scope of the SDK. You'll need to base your types off a common one in your world for your own use-cases. If we supported all types than the SDK would need to be the source of truth for several projects/plugins and it becomes a bit unmanageable and violates the separation of concerns principle.

I do believe we have an issue with exporting the `K8sResourceCommon` type right now, but it should be there.

config/dev.webpack.config.js
```diff
@@ -18,14 +18,14 @@ const calculateRemoteConfig = (remoteConfig) => {
...
   useProxy: true,
   env,
   appUrl: process.env.BETA ? '/beta/hac/app-studio' : '/hac/app-studio',
   standalone: Boolean(process.env.STANDALONE),
-  ...(process.env.INSIGHTS_CHROME && {
-    localChrome: process.env.INSIGHTS_CHROME,
-  }),
+  // ...(process.env.INSIGHTS_CHROME && {
+  //   localChrome: process.env.INSIGHTS_CHROME,
+  // }),
```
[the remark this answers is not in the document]
karelhala: Good point! I should remove the comments, this is usefull for local development.

config/plugins.js
```diff
@@ -1,16 +1,89 @@
...
 const plugins = [
-  fedModulePlugin({
-    root: resolve(__dirname, '../'),
-    debug: true,
-    exposes: {
-      // Application root
-      './RootApp': resolve(__dirname, '../src/AppEntry'),
-    },
-  }),
+  // new DynamicRemotePlugin(remotePluginOptions),
```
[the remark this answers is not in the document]
karelhala: Not anymore! Removed.

src/components/AddComponent/utils.ts
```diff
@@ -33,26 +30,10 @@ export const useAccessTokenBindingAuth = (name: string) => {
     if (!name || !loaded) return;
     if (binding.status?.phase === SPIAccessTokenBindingPhase.Injected) {
       setFieldValue('git.authSecret', binding.status.syncedObjectRef.name);
-      dispatch(
-        addNotification({
```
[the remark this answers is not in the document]
karelhala: Unfortunately not. We either have to tap in the hac-core redux or write custom one.

karelhala:
@rohitkrai03 @andrewballantyne I've started to use the newest webpack plugin and am looking if we could somehow bridge redux to your app so you can utilize the notifications utilities.

rohitkrai03:
> @rohitkrai03 @andrewballantyne I've started to use the newest webpack plugin and am looking if we could somehow bridge redux to your app so you can utilise the notifications utilities.

@karelhala Based on UX inputs we would like to remove the notification utilities anyways. So its okay to replace them with `console.log` statements for now.

config/plugins.js
```diff
@@ -1,17 +1,9 @@
 const { resolve } = require('path');
+const pckg = import('@openshift/dynamic-plugin-sdk-webpack');
```
andrewballantyne, line 2: Why Async?

src/dynamic-plugin-sdk.ts
```diff
@@ -3,11 +3,14 @@ import * as React from 'react';
 import * as _ from 'lodash';
 import { HttpError } from './shared/utils/error/http-error';
 import { useDeepCompareMemoize } from './shared';
+import {
+  k8sListResourceItems,
+  k8sGetResource,
+} from '@openshift/dynamic-plugin-sdk-utils';
```
andrewballantyne, line 9: Should be able to drop this file entirely. I was able to delete this variant in HAC-Core when I pulled alpha3

src/pages/TestK8s.tsx
```diff
@@ -1,18 +1,16 @@
...
   k8sPatchResource,
-  K8sResourceCommon,
   k8sUpdateResource,
-  // useK8sWatchResource,
-  // WatchK8sResource,
-} from '../dynamic-plugin-sdk';
+  // K8sResourceCommon,
+} from '@openshift/dynamic-plugin-sdk-utils';
+import { Button, PageSection, TextInput } from '@patternfly/react-core';
+import { K8sModel, K8sResourceCommon } from '../dynamic-plugin-sdk';
```
andrewballantyne, line 13: These are available now.
```suggestion
import { K8sModelCommon, K8sResourceCommon } from '@openshift/dynamic-plugin-sdk-utils';
```

src/pages/TestK8s.tsx
andrewballantyne: On second thought -- we can do this later if needed -- let us avoid this overhead for now.

andrewballantyne (changes requested):

config/plugins.js
karelhala: The webpack config uses commonJS (we can move it to esmodule if we want to, but that is probably outside of scope of this PR) and this plugin is ES module so we have to use `import` as a fallback

src/dynamic-plugin-sdk.ts
andrewballantyne: We should identify those and if needed bring those over to the SDK... we shouldn't keep this file, it was very much a hack.

config/plugins.js
andrewballantyne: Ah right -- that's unfortunate. Okay, thanks for the info.

src/dynamic-plugin-sdk.ts
andrewballantyne: They shouldn't need `commonFetch` either -- if they do need it they are not properly hooked into HAC-Core.

src/models/application.ts
```diff
@@ -1,13 +1,11 @@
...
 
-export const ApplicationModel: K8sModel = {
+export const ApplicationModel: K8sModelCommon = {
   apiGroup: 'appstudio.redhat.com',
   apiVersion: 'v1alpha1',
   kind: 'Application',
   plural: 'applications',
-  abbr: 'APP',
-  label: 'Application',
-  labelPlural: 'Applications',
```
andrewballantyne, line 10: Hmm, I wonder if these are needed fields @rohitkrai03 if the Plugin adds them. These are half-holdovers from OpenShift Console to use models to show details on pages. I'm not a huge fan of this model, but the alternative is probably some mixture map to values.

Do you guys have a need for ResourceIcons? What about pages that read models and show labels like a page header?

src/models/component.ts
```diff
@@ -1,13 +1,11 @@
...
+export const ComponentModel: K8sModelCommon = {
   apiGroup: 'appstudio.redhat.com',
   apiVersion: 'v1alpha1',
   kind: 'Component',
   plural: 'components',
-  abbr: 'COMP',
-  label: 'Component',
-  labelPlural: 'Components',
   namespaced: true,
 };
```
andrewballantyne, line 10: same here

src/models/component.ts
```diff
@@ -21,10 +19,7 @@ export const ComponentDetectionQueryModel = {
   apiGroup: 'appstudio.redhat.com',
   apiVersion: 'v1alpha1',
   kind: 'ComponentDetectionQuery',
-  abbr: 'CDQ',
   plural: 'componentdetectionqueries',
-  label: 'Component Detection Query',
-  labelPlural: 'Component Detection Queries',
```
andrewballantyne, line 27: and here

src/models/project.ts
```diff
@@ -1,11 +1,8 @@
...
+import { K8sModelCommon } from '@openshift/dynamic-plugin-sdk-utils';
 
-export const ProjectModel: K8sModel = {
+export const ProjectModel: K8sModelCommon = {
   apiVersion: 'v1',
   apiGroup: 'project.openshift.io',
   kind: 'Project',
-  abbr: 'PR',
-  label: 'Project',
-  labelPlural: 'Projects',
```
andrewballantyne, line 9: and here

src/models/service-provider-integration.ts
```diff
@@ -1,13 +1,11 @@
...
 
-export const SPIAccessTokenBindingModel: K8sModel = {
+export const SPIAccessTokenBindingModel: K8sModelCommon = {
   apiGroup: 'appstudio.redhat.com',
   apiVersion: 'v1beta1',
   kind: 'SPIAccessTokenBinding',
   plural: 'spiaccesstokenbindings',
-  abbr: 'ATB',
-  label: 'SPIAccessTokenBinding',
-  labelPlural: 'SPIAccessTokenBindings',
```
andrewballantyne, line 10: and here

andrewballantyne (commented):
Other uses to my previous comment

andrewballantyne (changes requested):
Only one more to my previous comments I noticed

src/shared/components/catalog/utils/catalog-utils.tsx
```diff
@@ -50,7 +51,7 @@ export const updateURLParams = (paramName: string, value: string | string[]) =>
   } else {
     params.delete(paramName);
   }
-  setURLParams(params);
+  // setURLParams(params);
```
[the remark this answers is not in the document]
karelhala: Good point! I've commented it to see what was causing the test fails. Uncommented it!

config/remotePlugin.js
```diff
@@ -0,0 +1,117 @@
...
+module.exports = {
+  entryCallbackSettings: {
+    name: 'loadPluginEntry',
+    pluginID: `${packageInfo.insights.appname}@${packageInfo.version}`,
+  },
+  pluginMetadata: {
+    name: packageInfo.name,
+    version: packageInfo.version,
+    exposedModules: {
+      AppEntry: resolve(__dirname, '../src/components/SamplesFlow.tsx'),
```
rohitkrai03, line 13:
```suggestion
      AppEntry: resolve(__dirname, '../src/components/AppFlow.tsx'),
```

rohitkrai03 (changes requested):
Tried running the PR locally. For some reason the requests to fetch applications is failing on the application list page.
Screenshot 2022-03-10 at 9 05 08 PM

rohitkrai03 (changes requested):
@karelhala A few more issues I noticed.
- I still see the issues with CSS and all the routes seem to be rendered inside a grey box (padding on all sides). 
- We have the `AppBanner` that says `Development Preview` which is supposed to be rendered on all app studio pages. This is gone now because we do not have a `App.tsx` anymore. Is there a way we can do that now in this plugin model?

karelhala:
@rohitkrai03 for the grey background, that's platform specific and should be present on all pages your UX should discuss this with wider audience, if you want I can give you point of contacts. As for the Dev preview, that shouldn't be too hard to add. The CSS might be a bit more tricky.

rohitkrai03:
> @rohitkrai03 for the grey background, that's platform specific and should be present on all pages your UX should discuss this with wider audience, if you want I can give you point of contacts. As for the Dev preview, that shouldn't be too hard to add. The CSS might be a bit more tricky.

I am okay with the overall grey background, but the issue is whatever components we are trying to render it gets rendered within some padding on all sides. That breaks the UX that we have, I think we should be able to use the full width of the content section without any extra padding on all sides.

For comparison take a look at the `console.redhat.com` page and the hac-dev page below - 

HAC Dev - 
Screenshot 2022-03-11 at 8 54 34 PM

Console.redhat.com (Notice the title section on top is rendered full width and doesn't have a padding.)
Screenshot 2022-03-11 at 8 54 44 PM

karelhala:
@rohitkrai03 I have figured out why the classes were not properly applied. Turns out the problem was previously some styling was applied in AppEntry, which is no longer used.

@andrewballantyne I've also updated hac-core to wrap the dynamic component with `section` instead of `Main` so we keep the styling same as before https://github.com/openshift/hac-core/pull/44

karelhala:
@rohitkrai03 updated the code, should be running now. The custom styles will be applied once the hac-core PR is merged

package.json
```diff
@@ -9,31 +9,32 @@
   "scripts": {
     "analyze": "NODE_ENV=production webpack --config config/prod.webpack.config.js --env analyze=true",
     "build": "NODE_ENV=production webpack --config config/prod.webpack.config.js",
-    "coverage": "jest --coverage --runInBand --detectOpenHandles --forceExit",
```
jrichter1, line 12: We've added these parameters specifically to avoid the tests hanging on openshift-ci. Removing them might reintroduce the issue.

karelhala: I can try adding them back, but when I try to run it locally with them it hangs on the very first test.

karelhala:
@rohitkrai03 The `codecov/patch` report showing 45% of diff is strange. I had to re-arrange a lot of code by using different imports. Looks like these lines were never tested and now it fails the task. Should I go hunting these lines and write tests for them?

rohitkrai03:
> @rohitkrai03 The `codecov/patch` report showing 45% of diff is strange. I had to re-arrange a lot of code by using different imports. Looks like these lines were never tested and now it fails the task. Should I go hunting these lines and write tests for them?

@karelhala I think it's okay to manually merge this if codecov is becoming an issue.

karelhala:
@rohitkrai03 mhh, yeah. There has been quite a few changes to routing, updated. Thank you for taking a look!

rohitkrai03 (changes requested):
@karelhala I found some more issues in the workflows - 
- Cannot view build logs for any of the components. Its always saying `No pipeline runs found` even though the pipeline run is available for that component.
- `Add Component` button doesn't take the user to create component form. It takes the user to `Create application` form.
- The `ComponentQueryDetection` doesn't work for any github URL.

https://user-images.githubusercontent.com/6041994/159274934-cab1fdb4-89dc-4ef4-b5a1-311bb8802f91.mov

karelhala:
> Add Component button doesn't take the user to create component form. It takes the user to Create application form.

This is the same on master as well, it leads to `create?application=${application}` shouldn't this lead to `create?component=${application}`?

rohitkrai03 (changes requested):
@karelhala Overall looks good. A few remaining issues that I found -
- The view logs modal is now in the content centre instead of app centre.
- Component delete operation does not work anymore. 

Screenshot 2022-03-21 at 10 53 16 PM

https://user-images.githubusercontent.com/6041994/159330789-1f42408f-9501-473a-873e-616c6e6c83bb.mov

karelhala:
@rohitkrai03 thank you! Fixed both of these issues, we don't have to append the modal to a component instead it's better to pass the class prefix to it so styles are applied. As for the delete operation for some reason the API version was not correct.

rohitkrai03 (approved):
Overall works fine. There are some CSS and layout issues that we'll follow up on another PR. @karelhala Thanks for the effort here for converting hac-dev to hac-core plugin.

/lgtm
/approve

PR #48, opened 2022-03-07
create Build logs viewer component for pipelineruns logs

sahil143:
## Fixes 

https://issues.redhat.com/browse/HAC-648

## Description

Add Build log viewer component to see pipeline run logs on UI

[TODO]:

- [ ] User Websocket/fetch polling to stream logs
- [ ] fix `hacDev` style scoping for modals

## Type of change

- [X] Feature
- [ ] Bugfix
- [ ] Build related changes

Kapture 2022-03-07 at 17 11 31

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


package.json
```diff
@@ -35,8 +35,12 @@
     "@redhat-cloud-services/frontend-components-notifications": "^3.2.2",
     "@redhat-cloud-services/frontend-components-utilities": "^3.2.4",
     "classnames": "^2.3.1",
+    "dayjs": "^1.10.8",
+    "file-saver": "1.3.x",
     "formik": "^2.2.9",
     "history": "4.x",
+    "i18next": "^19.8.3",
```
rohitkrai03, line 42: Where is this being used? Doesn't `react-i18next` give us enough utilities around translations?

rohitkrai03 (changes requested):

rohitkrai03 (changes requested):
@sahil143 I tried it locally. I can only see the step name as logs and not the full log. Also sometimes it is not able to fetch the resources and gets stuck in loading state.

https://user-images.githubusercontent.com/6041994/157901275-f9da691d-a22e-424c-b5b5-ca2c265f37f3.mov

package.json
sahil143: @rohitkrai03 this is for utilities. `i18next.t()`

src/shared/components/pipeline-run-logs/logs/LogsWrapperComponent.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  if (loaded && !error && resource.name === obj.metadata.name) {
+    resourceRef.current = obj;
+  }
+
+  return resourceRef.current ? (
+    <MultiStreamLogs {...props} resourceName={resource.name} resource={resourceRef.current} />
+  ) : (
+    <LoadingBox />
+  );
```
sahil143, line 26: @rohitkrai03 Loading of the pod is a bit slow because of the polling we do in useK8sWatchResource. when a task is selected new pod is fetched but useK8sWatchResource sends updates of the previously fetched resource that causes fluctuation on UI between the current task and previously selected task. Hence this logic was added to keep the pod resource consistent and to Send resource updates to MultiStreamLog component only when the `resource.name` matches the result from useK8sWatchResource.
Kapture 2022-03-11 at 23 00 31

rohitkrai03 (changes requested):
@sahil143 We might need an empty state for the log viewer when no pipeline run is available for a component. Right now it shows a loading indicator. Also we should increase the polling delay to improve the logs performance.

https://user-images.githubusercontent.com/6041994/158218658-eb46cb4e-6412-4a13-8420-bb6cfded7302.mov

PR #49, opened 2022-03-07
Milestone 3 UX refinement changes

rohitkrai03:
## Fixes 

- https://issues.redhat.com/browse/HAC-725

## Description

UX team wanted some changes to UX for Milestone 3. Below are the changes that the PR covers - 
- Removed workspace dropdown from create application form.
- Removed side panel from samples catalog.
- Added sticky footer to forms.

## Type of change

- [x] Other (please describe): UX related usability enhancements.

https://user-images.githubusercontent.com/6041994/157050569-670da1e6-934e-4720-9769-13926b400ab6.mov

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


christianvogt:
The sticky footer has drop shadow issues where the top shadow doesn't span the full width and the left and right shadows are showing:
image

rohitkrai03:
> The sticky footer has drop shadow issues where the top shadow doesn't span the full width and the left and right shadows are showing: image

The problem is with the layout we are using. The main page container has paddings on both sides which is why the footer is not full width. This would require a lot more refactoring, so created another issue to track it - https://issues.redhat.com/browse/HAC-794

rohitkrai03:
Merging the PR manually since the check for builds is not correctly setup as of now. All the other checks are passing on the PR.

PR #69, opened 2022-04-07
Use default entry callback settings

karelhala:
## Fixes 
Inaccessible UI due to error `Uncaught ReferenceError: loadPluginEntry is not defined`

## Description
There's a new usage of default entry callback available and it's not compatible with the old one. Let's use the first the new one so we don't break the UI.

## Type of change

- [x] Bugfix


rohitkrai03 (approved):
/approve
/lgtm

PR #72, opened 2022-04-12
Add useParams hook to calculate matching extension route

karelhala:
## Fixes 
Unable to use `useParams` from `react-router-dom`

## Description
Since hac-core has to use non-exact route to match any dynamic route it swallows the entire route and `useParams` from `react-router-dom` will not sent proper params even if the route matches. This PR introduces custom hook that goes trough all routes defined in `remotePlugin` and if matching route to current pathname is found return it. If there would be more than one match return just the first found (also break the reducer after finding the first route).

## Type of change

- [x] Feature


christianvogt:
@karelhala can you explain the usecase / problem in more detail and where this will be used as it hasn't been put to use in this PR.

karelhala:
@christianvogt this was requested by @rohitkrai03 on slack, the applications path is using query to identify current application. However it could be improved and use params in route to identify the app. This way the URL would be cleaner and could be chained with for instance detail fo component of app.

christianvogt:
Will followup on that thread.

I'd like to know why we don't render `Route` as react-router intends.

rohitkrai03:
@karelhala Should we close this one since the routes are getting rendered with react-router directly now?

PR #78, opened 2022-04-19
[HAC-998] Deploy to Ephemeral

BlakeHolifield:
* Deploys HAC and its dependences to a random ephemeral environment
* Sets up authentication in the central keycloak
* Registers the service for use in the dev sandbox cluster.

On every PR that's opened in this repo, it will deploy a frontend for testing. e2e tests with Cypress will come as next steps.


tmp/keycloak.py
```diff
@@ -0,0 +1,102 @@
...
+        "attributes": {
+            "firstName": "Ephemeral",
+            "lastName": "Environment",
+            "AccountID": "12345",
+            "AccountNumber": 12345,
+            "OrgID": "12345",
+            "IsInternal": "false",
+            "IsOrgAdmin": "true",
+            "IsActive": "true",
+            "Entitlements": '"insights": {"is_entitled": "true", "is_trial": "false"}'
```
karelhala, line 57: Do we need to send `insights` entitlements? I think it should be enough to just send empty object.

BlakeHolifield: I talked with Pete, and supporting this is going to be a bit of an effort to update hac-proxy. The user we're adding is already registered and we'd like to get this merged for e2e support.

tmp/hac-proxy.yaml
```diff
@@ -0,0 +1,108 @@
...
+      rules:
+        - host: ${HOSTNAME}
+          http:
+            paths:
+              - backend:
+                  service:
+                    name: hac-proxy
+                    port:
+                      number: 8080
+                path: /api/k8s
```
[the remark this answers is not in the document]
BlakeHolifield: I talked with Pete, and supporting this is going to be a bit of an effort to update hac-proxy. The user we're adding is already registered and we'd like to get this merged for e2e support.

psav (approved):

jerolimov (approved):
Thanks Pete for confirm this change.

/approve

PR #80, opened 2022-04-20
Check for completed status in CDQ

rohitkrai03:
## Fixes 
- https://issues.redhat.com/browse/HAC-1048

## Description

- The CDQ added a new `Processing` status in the resource. We need to stop looking for CDQ resource once the status is `Completed`.

## Type of change

- [x] Bugfix

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/AddComponent/utils.ts
+++ b/src/components/AddComponent/utils.ts
@@ -102,9 +102,14 @@ export const useComponentDetection = (
       if (loadError) {
         return loadError;
       }
-      if (loaded && !detectedComponents && cdq?.status?.conditions?.[0]?.status === 'True') {
+
+      const completeCondition = cdq?.status?.conditions?.find(
+        (condition) => condition.type === 'Completed',
+      );
+
+      if (loaded && !detectedComponents && completeCondition) {
         if (cdqName === cdq.metadata.name) {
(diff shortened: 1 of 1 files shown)


src/components/AddComponent/utils.ts
```diff
@@ -102,9 +102,9 @@ export const useComponentDetection = (
       if (loadError) {
         return loadError;
       }
-      if (loaded && !detectedComponents && cdq?.status?.conditions?.[0]?.status === 'True') {
```
rottencandy, line 105: I think this might break again if in the future CDQ gets another status field, what if we instead filter the conditions by type to get the correct condition?

rottencandy (commented):
We need to update the CDQ watch hook as well:
https://github.com/openshift/hac-dev/blob/2eaadb3b610609a49c99d7cff3842e1aaff00045/src/components/AddComponent/utils.ts#L105

src/components/AddComponent/utils.ts
rohitkrai03: Done.

rottencandy (approved):
/lgtm

PR #84, opened 2022-04-26
Update custom proxy so that it doesn't clash with k8s APIs

rohitkrai03:


divyanshiGupta:
Signup api is also called in handleSubmit of SignupView component. Please update the api url there as well.

divyanshiGupta (changes requested):
Signup api is also called in handleSubmit of SignupView component. Please update the api url there as well.

PR #87, opened 2022-04-27
Add error boundary to namespaced pages

rottencandy:
## Fixes 
https://issues.redhat.com/browse/HAC-1016

## Description

Adds an error boundary to NamespacedPage.

## Type of change

- [x] Feature
- [x] Bugfix

## Browser conformance: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/ErrorBoundary/ErrorBoundary.tsx
+++ b/src/components/ErrorBoundary/ErrorBoundary.tsx
@@ -1,0 +1,57 @@
+import React from 'react';
+import { ErrorBoundaryFallback } from './ErrorBoundaryFallback';
+
+export type ErrorBoundaryFallbackProps = {
+  errorMessage: string;
+  componentStack: string;
+  stack: string;
+  title: string;
+};
+
+type ErrorBoundaryProps = {
+  FallbackComponent?: React.ComponentType<ErrorBoundaryFallbackProps>;
+};
+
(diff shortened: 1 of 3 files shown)


src/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,57 @@
...
+  FallbackComponent?: React.ComponentType<ErrorBoundaryFallbackProps>;
+};
+
+type ErrorBoundaryState = {
+  hasError: boolean;
+  error: Error;
+  errorInfo: React.ErrorInfo;
+};
+
+export class ErrorBoundary extends React.Component<ErrorBoundaryProps, ErrorBoundaryState> {
```
rohitkrai03, line 21: Why is this written as class component?

rohitkrai03 (changes requested):

src/components/ErrorBoundary/ErrorBoundary.tsx
rottencandy: There is no functional component equivalent of `componentDidCatch` yet: https://stackoverflow.com/a/59932120/7683374
We'd have to go with class component, or use a library like https://github.com/bvaughn/react-error-boundary

PR #91, opened 2022-04-29
Fix: Use IMAGE instead of component name

BlakeHolifield:
To match feature parity with backed apps, we use `COMPONENT` as our app-interface folder and use the full `IMAGE` name.


florkbr:
/lgtm 

@rohitkrai03 @invincibleJai @jerolimov  for final approval

jerolimov (approved):
/approve

PR #97, opened 2022-05-11
Test Application Creation via Git source components

dheerajodha:
## Fixes 

Fixes: https://issues.redhat.com/browse/HAC-1291

## Description

Adds a UI test that automates the creation and deletion of the component of an application by using a Git URL as a source.

## Type of change

- [x] Other (please describe):
Adds UI test using cypress.

## How to test or reproduce?
Please read the readme in the integration tests folder.

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


integration-tests/support/pages/BuildApplicationWithGitSourcePage.ts
```diff
@@ -0,0 +1,43 @@
+import { createApplicationPagePO } from '../pageObjects/createApplication-po';
+import { buildApplicationWithGitSourcePagePO } from '../pageObjects/createApplication-po';
+import { AbstractWizardPage } from './AbstractWizardPage';
+export class BuildApplicationWithGitSourcePage extends AbstractWizardPage {
```
Katka92, line 4: What about merging this class with [https://github.com/openshift/hac-dev/blob/main/integration-tests/support/pages/AddComponentPage.ts](https://github.com/openshift/hac-dev/blob/main/integration-tests/support/pages/AddComponentPage.ts ) as it already contains some logic from the same page?

integration-tests/support/pageObjects/createApplication-po.ts
```diff
@@ -31,3 +31,11 @@ export const componentsListPagePO = {
   applicationName: 'pf-c-title pf-m-4xl hacDev-page__heading',
   items: '[pf-c-data-list__item]',
 };
+
+export const buildApplicationWithGitSourcePagePO = {
```
Katka92, line 35: Maybe worth merging with *addComponentPagePO* as it already contains some elements of this page.

integration-tests/support/pages/BuildApplicationWithGitSourcePage.ts
```diff
@@ -0,0 +1,43 @@
...
+  clickGitOptions() {
+    cy.contains(buildApplicationWithGitSourcePagePO.gitOptions).click();
+  }
+
+  setGitReference(gitReference: string) {
+    this.clearGitReference();
+    cy.get(buildApplicationWithGitSourcePagePO.gitReference).type(gitReference);
+  }
+
+  clearGitReference() {
```
Katka92, line 27: I'm not sure if we need this logic to be separated into an extra method as it contains the reference usable only with this particular element. Applies to other "clear" methods too.

Katka92 (changes requested):

integration-tests/tests/create-application-from-git-source.spec.ts
```diff
@@ -0,0 +1,44 @@
+// <reference types="cypress" />
+import { AddComponentPage } from '../support/pages/AddComponentPage';
+import { ApplicationComponentPage } from '../support/pages/ApplicationComponentsPage';
+import { ComponentSamplesPage } from '../support/pages/ComponentSamplesPage';
+import { CreateApplicationPage } from '../support/pages/CreateApplicationPage';
+import { BuildApplicationWithGitSourcePage } from '../support/pages/BuildApplicationWithGitSourcePage';
+import { Common } from '../utils/Common';
+
+describe('Create Application from Git Source', () => {
```
jrichter1, line 9: it's the component that is being created from git, not the application, so the terminology is a bit off

integration-tests/tests/create-application-from-git-source.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+  const applicationName = `test-app${new Date().getTime() / 1000}`;
+
+  it('Creating a Quarkus component', () => {
+    //set application name
+    const createApplicationPage = new CreateApplicationPage();
+    createApplicationPage.clickCreateApplication();
+    createApplicationPage.setApplicationName(applicationName);
+    createApplicationPage.clickNext();
```
jrichter1, line 17: creating the application would make more sense in the `before` hook

integration-tests/tests/create-application-from-git-source.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+  it('Delete java quarkus component', () => {
+    //Open components page
+    Common.openApplicationURL(applicationName);
+    //Review component page
+    const applicationPage = new ApplicationComponentPage();
+    applicationPage.deleteComponent('1-java-quarkus');
+
+    //Check if application does not exists
+    applicationPage.createdApplicationNotExists('1-java-quarkus');
+  });
```
jrichter1, line 43: we already have this exact test, so this looks more like a cleanup step, which should be in the `after` hook

integration-tests/support/pages/BuildApplicationWithGitSourcePage.ts
dheerajodha: Sure, I've deleted this file and moved the methods to the AddComponentPage class. Can you please take a look again and see if it looks good?

integration-tests/tests/create-application-from-git-source.spec.ts
dheerajodha: Makes sense :+1: thanks

integration-tests/tests/create-application-from-git-source.spec.ts
dheerajodha: done, thanks!

integration-tests/support/pages/BuildApplicationWithGitSourcePage.ts
Katka92: Nice, thanks :+1:

integration-tests/support/pageObjects/createApplication-po.ts
Katka92: Thanks!

integration-tests/support/pages/BuildApplicationWithGitSourcePage.ts
Katka92: Looks better to me, thanks. :slightly_smiling_face:

integration-tests/support/pageObjects/createApplication-po.ts
```diff
@@ -1,6 +1,12 @@
 export const addComponentPagePO = {
   samples: 'Start with a sample.',
   addComponent: '[data-testid="add-component"]',
+  enterSource: '[id="form-input-source-field"]',
+  validated: '[id="form-input-source-field-helper"]',
+  gitOptions: 'Git options',
+  gitReference: '[id="form-input-git-reference-field"]',
+  contextDir: '[id="form-input-git-contextDir-field"]',
+  next: '[data-test="submit-button"]'
```
rohitkrai03, line 9: Shouldn't we be consistent here. Some have `data-testid`, some have `data-test` and some `id`?

rohitkrai03 (changes requested):

integration-tests/support/pageObjects/createApplication-po.ts
dheerajodha: Sure, I've made some changes in the corresponding files of the components by adding a new `data-testid` field as it wasn't present before, and then updated the test accordingly.

rohitkrai03: @jrichter1 Do you remember what is the standard (`data-test` or `data-testid`) we use across console?

jrichter1: there is the legacy `data-test-id` and the current  is `data-test`

dheerajodha: so should I replace `data-testid` with `data-test`?

rohitkrai03 (approved):
/approve

PR #98, opened 2022-05-12
Fix namespace not being initialised into form context

rohitkrai03:
## Fixes 

- https://issues.redhat.com/browse/HAC-1167

## Description

- The namespace was being set in namespaced context and initialised into form context in application create page. But if we add component to already created application this step would not happen and namespace would not be initialized. 

## Type of change

- [x] Bugfix


jrichter1:
Looks like the e2e tests are indeed no longer failing. It would also be nice if the change did not decrease the code coverage.

rottencandy (approved):
/lgtm

PR #100, opened 2022-05-19
Pass git options to component detection hook

rottencandy:
## Fixes 
https://issues.redhat.com/browse/HAC-1285

## Description
When git options are entered, the callback is triggered which sets the field to validating, but since git options aren't used by the detection hook a new CDQ is not created. This causes the field to never get validated.

Solution:
- Do not hide git options section if any value is changed.
- Pass git options to detection hook so it creates a new CDQ when they change, even though we don't use git options currently.
- Debounce the creation of CDQ in the hook so that it doesn't try to create a new CDQ on every keystroke.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

https://user-images.githubusercontent.com/20013884/169376786-f7b0fc63-bcac-4680-adf4-9b461e865404.mp4

## How to test or reproduce?
Try to update git options after a repo is entered.

## Browser conformance: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/AddComponent/SourceSection.tsx
+++ b/src/components/AddComponent/SourceSection.tsx
@@ -45,18 +45,18 @@ export const SourceSection: React.FC<SourceSectionProps> = ({ onSamplesClick }) 
   const [detectedComponents, loadError] = useComponentDetection(
     sourceUrl,
     formState.application,
-    gitOptions.isMultiComponent,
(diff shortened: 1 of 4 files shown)


rohitkrai03 (approved):
/approve
/lgtm

PR #104, opened 2022-05-24
e2e tests: collect cypress logs

jrichter1:
into a file so we can see in detail what happens in CI runs

### DIFF
--- a/integration-tests/support/commands/index.ts
+++ b/integration-tests/support/commands/index.ts
@@ -6,3 +6,9 @@ import './hooks';
 Cypress.on('uncaught:exception', (err) => {
   return false;
 });
+
+// Add browser logs collector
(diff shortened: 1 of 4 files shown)


integration-tests/support/commands/index.ts
```diff
@@ -6,3 +6,9 @@ import './hooks';
 Cypress.on('uncaught:exception', (err) => {
   return false;
 });
+
+// Add browser logs collector
+const logOptions = {
+  enableExtendedCollector: true
+};
+require('cypress-terminal-report/src/installLogsCollector')(logOptions);
```
Katka92, line 14: No newline at the end.

Katka92 (approved):
Thanks for a great enhancement!

PR #105, opened 2022-05-25
Add QE as owners of integration-tests folder

jrichter1:


integration-tests/OWNERS
```diff
@@ -0,0 +1,5 @@
+reviewers:
+  - jkopriva
```
rohitkrai03, line 2: Shouldn't you also add approvers as reviewers?

rohitkrai03 (changes requested):

integration-tests/OWNERS
jrichter1: AFAIK approvers should also be able to review as per https://github.com/kubernetes/community/blob/master/community-membership.md#approver

PR #128, opened 2022-06-23
Fix review components page crash

rohitkrai03:
## Fixes 

- https://issues.redhat.com/browse/HAC-1625

## Description

- Null check was needed in case of git flow where container image is undefined.

## Type of change

- [x] Bugfix

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


rottencandy (approved):
/lgtm

PR #129, opened 2022-06-23
Upgrade react-router-dom to version 6

karelhala:
## Fixes 
Upgrades react-router-dom to version 6

## Description
Since platform would like to update to react router dom version 6 we should update as soon as possible to it as well. Hac-core already opened a PR to update the version so we should update the hac-dev as well.

## Type of change

- [x] Build related changes

## Checklist:

- [x] merge after https://github.com/openshift/hac-core/pull/80


rohitkrai03 (approved):
/lgtm
/approve

PR #136, opened 2022-07-04
Add Hacbs Application details view layout

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/HACBS-699

## Description

PR contains the application details view page layout
- Tabs with empty state
- Actions
- Components tab contains the component list as app studio

## Type of change

- [x] Feature

image

image

## Tests

```
 PASS  src/hacbs/components/ApplicationDetails/__tests__/DetailsPage.spec.tsx
  DetailsPage
    ✓ should render the DetailsPage (23 ms)
    ✓ should render the DetailsPage (5 ms)
    ✓ should not render the tabs if invalid values are passed (3 ms)
    ✓ should render the tabs (15 ms)
    ✓ should render the tabs (30 ms)
    ✓ should not render the actions if it is not passed (4 ms)
    ✓ should not render actions invalid values are passed to actions (5 ms)
```

```
 PASS  src/hacbs/components/ApplicationDetails/__tests__/HacbsApplicationDetails.spec.tsx
  HacbApplicationDetails
    ✓ should render spinner if application data is not loaded (21 ms)
    ✓ should render application display name if application data is loaded (47 ms)
```

## How to test?

* Visit Application details page
*  add `&hacbs=true` query param in the url to see the hacbs application details view

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

cc: @mfrances17

### DIFF
--- a/src/hacbs/components/ApplicationDetails/DetailsPage.tsx
+++ b/src/hacbs/components/ApplicationDetails/DetailsPage.tsx
@@ -1,0 +1,158 @@
+import * as React from 'react';
+import {
+  Dropdown,
+  DropdownItem,
+  DropdownItemProps,
+  DropdownSeparator,
+  DropdownToggle,
+  Flex,
+  FlexItem,
+  PageGroup,
+  PageSection,
+  PageSectionVariants,
+  Tab,
+  Tabs,
+  TabTitleText,
+  Text,
+  TextContent,
+} from '@patternfly/react-core';
+import { CaretDownIcon } from '@patternfly/react-icons/dist/esm/icons/caret-down-icon';
+import cx from 'classnames';
+import BreadCrumbs from '../../../shared/components/breadcrumbs/BreadCrumbs';
+import { getQueryArgument, removeQueryArguments, setQueryArgument } from '../../../shared/utils';
+
+import './DetailsPage.scss';
+
+type Action = { type?: string; key: string; label: string } & DropdownItemProps;
+type DetailsPageTabProps = {
+  key: string;
+  label: string;
+  component: React.ReactNode;
+  href?: string;
+  isDisabled?: true;
+  className?: string;
+  isFilled?: boolean;
+};
+
+type DetailsPageProps = {
+  title: string;
+  children?: React.ReactNode;
+  footer?: React.ReactNode;
+  description?: React.ReactNode;
+  breadcrumbs?: { name: string; path: string }[];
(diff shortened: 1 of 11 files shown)


src/hacbs/components/ApplicationDetails/DetailsPage.tsx
```diff
@@ -0,0 +1,151 @@
...
+  title,
+  footer,
+  description,
+  breadcrumbs,
+  actions = [],
+  tabs = [],
+  onTabSelect,
+}) => {
+  const tabMatched =
+    tabs?.find((t) => t.key === getQueryArgument('activeTab'))?.key || tabs?.[0]?.key;
```
christianvogt, line 53: You should be reading from react router for query changes to ensure reactivity. Right now the re-render depends on a higher up component causing this child to be re-rendered. But it's possible that this could break by using a pure component somewhere in between.

That being said, I'm not sure having the active tab as a query param is correct. For now it works though. We should talk with @rohitkrai03 to align on when to use query params vs path segments as we also use query params for the resource name.

src/hacbs/components/ApplicationDetails/HacbsApplicationDetails.tsx
```diff
@@ -0,0 +1,120 @@
...
+            key: 'add-component',
+            label: 'Add Component',
+            onClick: () => {
+              navigate(`/app-studio/import?application=${applicationName}`);
+            },
+          },
+          {
+            key: 'add-environment',
+            label: 'Add Environement',
+            href: '',
```
christianvogt, line 61: Always good to have an href so that a user can copy the link or open in a new tab.

src/hacbs/components/ApplicationDetails/tabs/ComponentsTab.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  const componentList = (
+    <ComponentListView applicationName={applicationName} components={filteredComponents} />
+  );
+  const loading = (
+    <Bullseye>
+      <Spinner />
+    </Bullseye>
+  );
+
```
christianvogt, line 45: nit: the spinner appears at the top of the content area
image

karthikjeeyar: Fixed it.

PR #138, opened 2022-07-06
Application details

sahil143:
## Fixes 

https://issues.redhat.com/browse/HAC-1606

## Description

Redesign application details view

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

image

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


src/hooks/usePipelineRunsForApplication.ts
```diff
@@ -0,0 +1,97 @@
...
+    namespace,
+    isList: true,
+  });
+  return React.useMemo(() => {
+    if (prLoaded && !prError) {
+      return pipelineRuns
+        ?.filter?.(
+          (plr) =>
+            plr.metadata.labels[BUILD_COMPONENT_LABEL] === component.metadata.name &&
+            plr.metadata.labels[BUILD_APPLICATION_LABEL] === component.spec.application,
```
rottencandy, line 25: Is it possible to use matchLabel to filter in the watch hook itself?

rottencandy:
If I open the delete app modal from details page and click cancel, it redirects to applicaiton list page.

Pretty sure this happened previously too, but we didn't notice since there was to delete button in details page.

src/hooks/usePipelineRunsForApplication.ts
rohitkrai03: Yes, it should be possible to use label selectors now. Core team fixed that bug.

sahil143: It doesn't seem to be working If I add matchLabels it returns empty array

src/components/modal/ModalProvider.tsx
```diff
@@ -2,30 +2,36 @@ import React from 'react';
...
-type ModalContextType = (launcher: ModalLauncher) => void;
+type ModalContextType<T = {}> = (launcher: ModalLauncher) => { closed?: Promise<T> };
 
-const ModalContext = React.createContext<ModalContextType>(() => {});
+const ModalContext = React.createContext(
+  // eslint-disable-next-line
+  <T extends {}>(launcher: ModalLauncher): { closed?: Promise<T> } => ({
+    closed: Promise.resolve<T>({} as T),
+  }),
+);
```
rottencandy, line 12: Would it be possible to specify type param and initializing with null?
```ts
const ModalContext =
  React.createContext<<T extends {}>(launcher: ModalLauncher) => { closed?: Promise<T> }>(null);

```

src/components/modal/ModalProvider.tsx
```diff
@@ -2,30 +2,36 @@ import React from 'react';
...
 export const ModalProvider: React.FC<ModalProviderProps> = ({ children }) => {
   const [modal, setModal] = React.useState<React.ReactElement>();
 
-  const handleClose = React.useCallback(() => {
-    setModal(null);
-  }, []);
-
-  const showModal = React.useCallback(
-    (launcherCallback) => {
+  const showModal = React.useCallback(<T extends {}>(launcherCallback): { closed?: Promise<T> } => {
```
rottencandy, line 24: And if we extract the type from context above, we can use it for the callback too.

src/components/modal/ModalProvider.tsx
```diff
@@ -2,30 +2,30 @@ import React from 'react';
...
```
rottencandy, line 18: Nit:
```suggestion
  const showModal = React.useCallback<ModalContextType>((launcherCallback) => {
```

rottencandy (approved):
/lgtm

integration-tests/utils/Common.ts
```diff
@@ -15,9 +15,9 @@ export class Common {
   }
 
   static openApplicationURL(applicationName: string) {
-    Common.openURL(
-      `${Cypress.env('HAC_BASE_URL')}/applications?name=${applicationName.replace('.', '-')}`,
-    );
+    if(!cy.contains('h1', applicationName, { timeout: 90000 })) {
```
jrichter1, line 18: AFAIK `cy.contains` doesn't actually produce a falsey result when nothing is found, it would just throw an error after the timeout runs out

A safer solution would be to check if the current url points to the appropriate application
 `.../applications?name=${applicationName}`

integration-tests/utils/Applications.ts
```diff
@@ -4,7 +4,7 @@ import { Common } from './Common';
 
 export class Applications {
   static deleteApplication(applicationName: string) {
-    Common.openAppStudioBaseURL();    
+    cy.get('[data-ouia-component-id="Applications"] > a').click();
```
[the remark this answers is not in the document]
jrichter1: also why not keep using `openAppStudioBaseURL` when you are updating it anyway?

rohitkrai03 (approved):
/approve
/lgtm

integration-tests/utils/Applications.ts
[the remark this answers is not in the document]
sahil143: because opening the url reloads the page which take a lot of time and sometime tests timeouts while waiting for the page to load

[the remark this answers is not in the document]
jrichter1: but you just changed the function to do the exact same as this line

PR #141, opened 2022-07-08
Add getting started cards & modals for onboarding experience

rottencandy:
## Fixes 
https://issues.redhat.com/browse/HAC-1610

## Description
Getting started cards & modals that use local storage to determine if a user is visiting for the first time.

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

## Browser conformance: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


src/pages/ApplicationsPage.tsx
```diff
@@ -1,21 +1,54 @@
...
+const GETTING_STARTED_CARD_KEY = 'application-list-getting-started-card';
 
 const ApplicationsPage = () => {
   useQuickstartCloseOnUnmount();
   const applicationName = getQueryArgument('name');
   const hacbs = JSON.parse(getQueryArgument('hacbs'));
 
+  const { namespace } = useNamespace();
+
+  const [applications] = useK8sWatchResource<ApplicationKind[]>({
```
rohitkrai03, line 28: If you're fetching all the applications here then maybe send it to the `ApplicationListView` as well.

src/pages/ApplicationsPage.tsx
```diff
@@ -1,21 +1,54 @@
...
+    groupVersionKind: ApplicationGroupVersionKind,
+    namespace,
+    isList: true,
+  });
+
+  if (!applications) {
```
rohitkrai03, line 34: Use loading variable from the hook. We don't want to show spinner when there's no application.

src/pages/ApplicationsPage.tsx
```diff
@@ -1,21 +1,54 @@
...
+import NamespacedPage, { useNamespace } from '../components/NamespacedPage/NamespacedPage';
 import PageLayout from '../components/PageLayout/PageLayout';
 import HacbsApplicationDetails from '../hacbs/components/ApplicationDetails/HacbsApplicationDetails';
 import { useQuickstartCloseOnUnmount } from '../hooks/useQuickstartCloseOnUnmount';
+import imageUrl from '../imgs/getting-started-illustration.svg';
+import { ApplicationGroupVersionKind } from '../models';
 import { getQueryArgument } from '../shared/utils';
+import { ApplicationKind } from '../types';
+
+const GETTING_STARTED_MODAL_KEY = 'application-list-getting-started-modal';
```
rohitkrai03, line 18: Maybe use just one key object for all of getting started cards?
```js
const GETTING_STARTED = {
  'APPLICATION_LIST': true,
}
```

rohitkrai03 (changes requested):

integration-tests/tests/create-application-from-git-source.spec.ts
```diff
@@ -73,26 +75,35 @@ describe('Create Component from Public Git Source', () => {
     it('Create Application', () => {
       componentPage.createApplication();
       applicationDetailPage.createdComponentExists(componentName, applicationName);
+      Applications.dismissOnboardingModal();
     });
 
     it('Check Component Build Log', () => {
+      Applications.dismissOnboardingModal();
```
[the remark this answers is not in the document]
rottencandy: Cypress seems to reset local storage after every single test case, but I can load the key so it doesn't show up if that's okay.

src/components/GettingStartedCard/GettingStartedCard.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const keys = typeof storageKeys === 'object' ? storageKeys : {};
+  const isDismissed = keys[localStorageKey];
+
+  return (
+    !isDismissed && (
+      <Card style={{ margin: '15px' }}>
+        <Split>
+          <SplitItem>
+            <img src={imgSrc} width="200" />
```
rhamilto, line 40: Make `imgSrc` optional and only include the parent `<SplitItem>` if it exists so the card can be used without an image?

rhamilto: Probably don't want to hard code `width`.

rhamilto: add `alt` for accessibility.

src/components/GettingStartedCard/GettingStartedCard.tsx
```diff
@@ -0,0 +1,61 @@
...
+}) => {
+  const [storageKeys, setStorageKeys] =
+    useLocalStorage<{ [key: string]: boolean }>(LOCAL_STORAGE_KEY);
+
+  const keys = typeof storageKeys === 'object' ? storageKeys : {};
+  const isDismissed = keys[localStorageKey];
+
+  return (
+    !isDismissed && (
+      <Card style={{ margin: '15px' }}>
```
rhamilto, line 37: Remove the inline style and wrap the card with a `<PageSection>` to control the margins?  I think we want the card's context to control the spacing.

src/components/GettingStartedCard/GettingStartedCard.tsx
```diff
@@ -0,0 +1,61 @@
...
+        <Split>
+          <SplitItem>
```
rhamilto, line 39: Probably want to the ability to set `className` here in order to control spacing around the image and display of the image.

for example, `<SplitItem className="pf-u-my-lg pf-u-ml-lg pf-u-display-none pf-u-display-block-on-md">`

rhamilto (commented):
Nice work.  A few comments based on my attempt to do something similar in https://github.com/openshift/hac-infra/pull/10.

src/pages/ApplicationsPage.tsx
rottencandy: Updated to use single object.

itsptk:
fyi @openshift/team-hacdev-ux 

@rottencandy feel free to tag the ux team so we can check out these PRs, thanks a lot

src/components/ApplicationListView/ApplicationListView.tsx
```diff
@@ -1,44 +1,27 @@
...
-      <Bullseye>
-        <Spinner />
-      </Bullseye>
-    );
-  }
+type ApplicationViewProps = {
+  applications: ApplicationKind[];
+};
 
+const ApplicationView: React.FC<ApplicationViewProps> = ({ applications }) => {
```
rohitkrai03, line 24:
```suggestion
const ApplicationListView: React.FC<ApplicationViewProps> = ({ applications }) => {
```

src/pages/ApplicationsPage.tsx
```diff
@@ -1,23 +1,59 @@
...
+const GETTING_STARTED_CARD_KEY = 'application-list-getting-started-card';
 
 const ApplicationsPage = () => {
   useQuickstartCloseOnUnmount();
   const applicationName = getQueryArgument('name');
   const [hacbs] = useFeatureFlag(HACBS_FLAG);
 
+  const [namespace, namespaceLoaded] = useActiveNamespace();
```
rohitkrai03, line 29: How is this going to work? The namespace context is getting set in this component only.

rohitkrai03 (changes requested):

src/pages/ApplicationsPage.tsx
rohitkrai03: It would also mean that the first time the page is loaded, namespace will be empty string and the resource fetch hook would make a call with `''` as namespace.

src/pages/ApplicationsPage.tsx
```diff
@@ -30,23 +42,7 @@ const ApplicationsPage = () => {
           )}
         </React.Fragment>
       ) : (
-        <React.Fragment>
-          <Helmet>
```
rohitkrai03, line 34: This could still be here just to be consistent. Just moving the paglayout stuff would work.

rohitkrai03 (changes requested):

src/pages/ApplicationsPage.tsx
rottencandy: Fixed.

PR #146, opened 2022-07-15
Allow clicking on "Applications" breadcrumb

dheerajodha:
## Description

This PR introduces the ability to click on the "Applications" breadcrumb (shown below)
While working on our tests, this can be further extended to click on other breadcrumbs as per our needs.

_Why do we need this?_
Because in our tests, to visit the "Applications" page (`hac/app-studio/` or `hac/app-studio/applications`) on AppStudio we are currently [using this](https://github.com/openshift/hac-dev/blob/da444ef22c2cf582ba33f37b256e505d7735817e/integration-tests/utils/Common.ts#L4). And it takes a LOT of time to load the page, which leads to test failures due to timeout. A better way to reach the same page is to reach it by just clicking the "Applications" breadcrumb on top.

image

## How to test or reproduce?

Add `Common.clickBreadcrumbs(breadcrumbs.applications);` as one of the steps in your tests inside the "integration-tests/tests" folder

### DIFF
--- a/integration-tests/support/pageObjects/global-po.ts
+++ b/integration-tests/support/pageObjects/global-po.ts
@@ -36,3 +36,7 @@ export const actions = {
   deleteModalInput: 'input[name*="resourceName"]',
   deleteModalButton: 'button[data-testid="delete-resource"]',
 }
+
+export const breadcrumbs = {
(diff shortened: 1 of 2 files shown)


dheerajodha:
Closing this PR and will put up a new one that has tests in it which use the function introduced in this PR.

PR #177, opened 2022-08-10
use context provider for Namespace

sahil143:
## Fixes 

https://issues.redhat.com/browse/HAC-1980

## Description

use context provider extension to get the namespace

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

No UI changes

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/config/remotePlugin.js
+++ b/config/remotePlugin.js
@@ -81,6 +81,7 @@ module.exports = {
       ComponentSettings: resolve(__dirname, '../src/pages/ComponentSettingsPage'),
       WorkspaceSettings: resolve(__dirname, '../src/pages/WorkspaceSettingsPage'),
       CreateEnvironment: resolve(__dirname, '../src/pages/CreateEnvironmentPage'),
+      NamespaceContext: resolve(__dirname, '../src/utils/namespace-context-utils'),
     },
   },
   extensions: [
@@ -241,6 +242,17 @@ module.exports = {
         exact: true,
         component: {
           $codeRef: 'CreateEnvironment',
+        },
+      },
+    },
+    {
+      type: 'core.context-provider',
+      properties: {
+        provider: {
+          $codeRef: 'NamespaceContext.NamespaceProvider',
+        },
+        useValueHook: {
+          $codeRef: 'NamespaceContext.useActiveNamespace',
         },
       },
     },
--- a/src/components/NamespacedPage/NamespacedPage.tsx
+++ b/src/components/NamespacedPage/NamespacedPage.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { Bullseye, Spinner } from '@patternfly/react-core';
-import { useActiveNamespace } from '../../hooks';
 import { UserSignupStatus, useSignupStatus } from '../../hooks/useSignupStatus';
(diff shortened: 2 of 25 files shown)


config/remotePlugin.js
```diff
@@ -81,6 +81,10 @@ module.exports = {
       ComponentSettings: resolve(__dirname, '../src/pages/ComponentSettingsPage'),
       WorkspaceSettings: resolve(__dirname, '../src/pages/WorkspaceSettingsPage'),
       CreateEnvironment: resolve(__dirname, '../src/pages/CreateEnvironmentPage'),
+      NamespaceContext: resolve(
+        __dirname,
+        '../src/components/NamespacedPage/namespace-context-provider',
```
christianvogt, line 86: I think we ought to move `useNamespace` and this context to the utils dir and out of the NamespacedPage component / dir.

src/components/NamespacedPage/NamespacedPage.tsx
```diff
@@ -9,23 +8,21 @@ import SignupView from '../Signup/SignupView';
 
 import './NamespacedPage.scss';
 
-type NamespaceContextData = {
-  namespace: string;
-};
+type NamespaceContextData = [string, boolean];
```
rohitkrai03, line 11: Does it have to be an array of string and boolean like that?

src/components/NamespacedPage/NamespacedPage.tsx
```diff
@@ -9,23 +8,21 @@ import SignupView from '../Signup/SignupView';
...
+
+const NamespaceContext = React.createContext<NamespaceContextData>(['', false]);
 
-export const NamespaceContext = React.createContext<NamespaceContextData>({
-  namespace: '',
-});
+export const NamespaceProvider = NamespaceContext.Provider;
 
-export const useNamespace = () => React.useContext(NamespaceContext);
+export const useNamespace = () => React.useContext(NamespaceContext)[0];
```
rohitkrai03, line 17: If you have the information of loaded or not maybe you should use it here in the hook somehow. What happens when someone tries to get the namespace from this hook and its not loaded yet?

src/components/NamespacedPage/NamespacedPage.tsx
sahil143: No, It doesn't. It can be an object.

src/components/NamespacedPage/NamespacedPage.tsx
sahil143: Our NamespacedPage Component handles this case. if namespace is not loaded we show a spinner. That's why it returns namespace string only.

src/components/NamespacedPage/NamespacedPage.tsx
rohitkrai03: Maybe having it as an object would be better in that case. Something like `{ namespace: string, loaded: boolean }`?

rohitkrai03 (approved):
/approve

sahil143:
image
image
image

PR #178, opened 2022-08-10
Show validation errors on application/component creation 

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/HACBS-985
https://issues.redhat.com/browse/HACBS-986

## Description

1. API errors are now surfaced in the form, so users can make changes and submit.
2. Fixed Add component issue.

## Type of change

- [x] Bugfix

Application creation:
image

Component Creation:

`edit` mode will be automatically enabled in case of errors in a component name field.

image

Integration Test:
image

## How to test or reproduce?

Go to Create application flow
1. Try to use the existing application/component names.

## Tests
```
  Submit Utils
  onApplicationSubmit
    ✓ should throw application name already exists error (4 ms)
    ✓ should throw same message from api response (1 ms)
    ✓ should create the application (1 ms)
  onComponentsSubmit
    ✓ should throw component name already exists error (1 ms)
    ✓ should throw component name already exists error when multiple components is passed (1 ms)
    ✓ should not throw error if the error object name mismatches with provided components (1 ms)
    ✓ should throw same message from api response (1 ms)
    ✓ should create the component

 ```
 ```
 EditableLabelField
    ✓ should set the component name field to editable mode and show the inline error message (16 ms)
   ``` 
 
## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

### CODE REVIEW

**Comment by karthikjeeyar** [comment]
cc: @christianvogt @dheerajodha

File: src/hacbs/components/ImportForm/utils/submit-utils.ts [hunk 1]
```diff
@@ -0,0 +1,30 @@
...
+};
+
+export const onApplicationSubmit = async ({ application, namespace }: FormValues, formikBag) => {
+  try {
+    await createApplication(application, namespace, true);
+    const applicationData = await createApplication(application, namespace, false);
+    formikBag.setFieldValue('applicationData', applicationData);
+  } catch (error) {
+    const message = error.code === 409 ? 'Application name already exists.' : error.message;
+    formikBag.setFieldError('application', message);
```
christianvogt, line 27: App studio sets the form status instead of a field error.
Also the status is set for component creation instead of the field error.

We should be consistent. So which one?

src/hacbs/components/ImportForm/utils/__tests__/submit-utils-spec.ts
```diff
@@ -0,0 +1,150 @@
...
+  setStatus: jest.fn(),
+  setFieldValue: jest.fn(),
+  setFieldError: jest.fn(),
+} as any;
+
+afterEach(() => {
+  jest.clearAllMocks();
+});
+
+describe('Submit Utils: onApplicationSubmit', () => {
```
christianvogt, line 25: Nit:
```suggestion
describe('submit-utils', () => {
  describe('onApplicationSubmit', () => {
```

or

```suggestion
describe('submit-utils#onApplicationSubmit', () => {
```

src/hacbs/components/ImportForm/utils/__tests__/application-data.ts
```diff
@@ -0,0 +1,38 @@
...
+            'f:displayName': {},
+          },
+        },
+        manager: 'Mozilla',
+        operation: 'Update',
+        time: '2022-08-09T18:52:59Z',
+      },
+    ],
+    name: 'my-application-3',
+    namespace: 'karthik-jk',
```
christianvogt, line 25:
```suggestion
    namespace: 'test',
```

karthikjeeyar: moved to __data__ folder and udpated it.

src/hacbs/components/ImportForm/utils/submit-utils.ts
```diff
@@ -0,0 +1,30 @@
...
+  try {
+    await createComponents(components, applicationData.metadata.name, namespace, secret, true);
+  } catch (error) {
+    const message = error.code === 409 ? 'Component name already exists.' : error.message;
+    formikBag.setStatus({ submitError: message });
+    throw error;
+  }
+};
+
+export const onApplicationSubmit = async ({ application, namespace }: FormValues, formikBag) => {
```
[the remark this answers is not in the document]
karthikjeeyar: done

src/hacbs/components/ImportForm/utils/submit-utils.ts
```diff
@@ -0,0 +1,33 @@
...
+import { ImportFormValues } from '../../../../components/ImportForm/utils/types';
+import { createApplication } from '../../../../utils/create-utils';
+import { FormValues } from '../types';
+
+export const onComponentsSubmit = async (
+  { components, applicationData, namespace, secret }: FormValues,
+  formikBag: FormikHelpers<ImportFormValues>,
+) => {
+  try {
+    await createComponents(components, applicationData.metadata.name, namespace, secret, true);
```
christianvogt, line 12: Components aren't created with dry run. But we should first perform a dry run followed by the real run.
See how app-studio does this in `submit-utils#createResources`

src/hacbs/components/ImportForm/utils/submit-utils.ts
beaumorley: Thanks for raising this @christianvogt 

The inline alert for the Application name is correct. We should follow that convention for the whole flow since the user needs guidance in what they need to fix. For the Component name error, I suggest we also use an inline alert and put the field into edit mode when the error is returned. So the user does not even have to click the pencil.

src/hacbs/components/ImportForm/utils/submit-utils.ts
karthikjeeyar: Updated it.

beaumorley:
@christianvogt @karthikjeeyar  Also wondering for the "Component already exists" scenario, are we auto incrementing the autogenerated name? Or do we always just use the runtime to generate so if they add two things that use the same runtime they will get the "already exists" error?

karthikjeeyar:
> @christianvogt @karthikjeeyar Also wondering for the "Component already exists" scenario, are we auto incrementing the autogenerated name? Or do we always just use the runtime to generate so if they add two things that use the same runtime they will get the "already exists" error?

@beaumorley The runtime is always being used as the component name and it throws the error when we try to use that runtime for the second time.

karthikjeeyar:
@beaumorley I have updated the PR, in case of error the component name field will be put into edit mode. PTAL a look at the  updated screenshots.

karthikjeeyar:
image

PR #169, opened 2022-08-05
Avoid creating binding when provider is unknown

rottencandy:
## Fixes 
https://issues.redhat.com/browse/HAC-1992

## Description
Avoid creating binding and show error message when a repo from unknown provider is used.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## How to test or reproduce?
Enter an unknown provider url.

## Browser conformance: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

PR #193, opened 2022-08-19
wait for access check to have a status

christianvogt:
## Fixes 
Fixes an issue related to https://github.com/openshift/hac-dev/pull/169

## Description

It was observed that entering a valid Git URL for a component may result in the error `This provider is not supported`:
image

The cause of this is that the SPI access check resource is fetched and considered `loaded` even though the controller may not have processed the check yet and therefore there is no `status` in the resource. At some later point in time, the controller will have processed the check and added a valid `status`. The fix here is to continue to wait until a `status` is present in the resource. We return `loaded = false` to the caller until the `status` appears. 

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

## How to test or reproduce?

To reproduce this, go to the import wizard and enter a Git URL for a component. If it immediately goes to validated, continue to change the git url until this error appears.

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

cc @rohitkrai03 @rottencandy

### DIFF
--- a/src/components/ImportForm/utils/__tests__/auth-utils.spec.ts
+++ b/src/components/ImportForm/utils/__tests__/auth-utils.spec.ts
@@ -109,7 +109,8 @@ describe('Auth Utils: useAccessCheck', () => {
     expect(result.current[1]).toBe(false);
   });
 
-  it('should return loaded true if source is present and loaded is true', async () => {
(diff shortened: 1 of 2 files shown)


rohitkrai03 (approved):
/lgtm

PR #205, opened 2022-08-29
Pipelinerun actions

mfrances17:
## Fixes 
Fixes https://issues.redhat.com/browse/HACBS-990

## Description
Stop and delete commands added. Rerun is still TBD.

## Type of change

- [x] Feature

## How to test or reproduce?
Go to the pipelinerun list page, select the kabob on a particular pipelinerun row and select Stop (to cancel a running PLR) or Delete (to bring up a modal and eventually delete the PLR).

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/hacbs/components/PipelineRunListView/pipelinerun-actions.tsx
+++ b/src/hacbs/components/PipelineRunListView/pipelinerun-actions.tsx
@@ -1,0 +1,33 @@
+import { k8sPatchResource } from '@openshift/dynamic-plugin-sdk-utils';
+import { pipelineRunFilterReducer } from '../../../shared/';
+import { Action } from '../../../shared/components/action-menu/types';
+import { PipelineRunModel } from '../../models/pipelineruns';
+import { PipelineRunKind } from '../../types';
+
+export const usePipelinerunActions = (pipelineRun: PipelineRunKind): Action[] => {
+  const pipelineRunStop = () => {
+    k8sPatchResource({
+      model: PipelineRunModel,
+      queryOptions: {
+        name: pipelineRun.metadata.name,
+        ns: pipelineRun.metadata.namespace,
+      },
+      patches: [
+        {
+          op: 'replace',
+          path: `/spec/status`,
+          value: 'StoppedRunFinally',
+        },
+      ],
+    });
+  };
+
+  return [
+    {
+      cta: () => pipelineRunStop(),
+      id: 'pipelinerun-stop',
+      label: 'Stop',
+      disabled: !(pipelineRunFilterReducer(pipelineRun) === 'Running'),
+    },
+  ];
+};
--- a/src/hacbs/components/PipelineRunListView/pipelinerun-actions.tsx
+++ b/src/hacbs/components/PipelineRunListView/pipelinerun-actions.tsx
@@ -16,7 +16,7 @@ export const usePipelinerunActions = (pipelineRun: PipelineRunKind): Action[] =>
         {
           op: 'replace',
           path: `/spec/status`,
-          value: 'StoppedRunFinally',
+          value: 'pipelineRunCancelled',
         },
       ],
     });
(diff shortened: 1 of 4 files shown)


christianvogt:
@mfrances17 I wasn't aware that there was supposed to be a delete action on a pipeline run. Could you share a link to this.

src/hacbs/components/Modals/resource-modals.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { PipelineRunKind } from '../../types';
+
+export const pipelineRunDeleteModal = (pipelineRun: PipelineRunKind) =>
+  createDeleteModalLauncher(pipelineRun.kind)({
+    obj: pipelineRun,
+    model: PipelineRunModel,
+    description: (
+      <>
+        The pipelinerun <strong>{pipelineRun.metadata.name}</strong> will be deleted from the
+        pipelinerun view and all the environments it is currently deployed.
```
christianvogt, line 13: @mfrances17 I don't think we want to delete pipeline runs. Nor do I believe that deletion of a pipeline run would have such cascading side effects as to remove a deployed snapshot from an environment.

christianvogt: The attachment you referred to was my screen grab from OpenShift Console where I was inquiring about the cancel action.

That's fine to log a new story for rerun and do it later.

mfrances17: i've removed Delete and will create a new story for Rerun.

src/hacbs/components/PipelineRunListView/pipelinerun-actions.tsx
```diff
@@ -0,0 +1,31 @@
...
+      model: PipelineRunModel,
+      queryOptions: {
+        name: pipelineRun.metadata.name,
+        ns: pipelineRun.metadata.namespace,
+      },
+      patches: [
+        {
+          op: 'replace',
+          path: `/spec/status`,
+          value: 'PipelineRunCancelled',
```
christianvogt, line 18: dev console uses `StoppedRunFinally`:
https://github.com/openshift/console/blob/master/frontend/packages/pipelines-plugin/src/utils/pipeline-actions.tsx#L180-L211

src/hacbs/components/PipelineRunListView/pipelinerun-actions.tsx
```diff
@@ -0,0 +1,31 @@
...
+        },
+      ],
+    });
+  };
+
+  return [
+    {
+      cta: () => pipelineRunStop(),
```
christianvogt, line 26: You're going to want to disable this action for any pipeline that isn't currently running. 
See https://github.com/openshift/console/blob/master/frontend/packages/pipelines-plugin/src/utils/pipeline-actions.tsx#L180-L211

src/hacbs/types/pipelineRun.ts
```diff
@@ -142,7 +142,7 @@ export type PipelineRunStatus = {
 
 export type PipelineRunKind = K8sResourceCommon & {
   spec: {
-    pipelineRef?: { name: string; bundle?: string };
+    pipelineRef?: { name: string };
```
karthikjeeyar, line 145: Please revert this change as `Tekton` now supports [bundle](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#tekton-bundles) field to support remotely defined pipelines.
```suggestion
    pipelineRef?: { name: string, bundle?: string };
```

src/hacbs/components/PipelineRunListView/pipelinerun-actions.tsx
karthikjeeyar: Yes, `PipelineRunCancelled` is deprecated.

mfrances17: thanks, updated to use `StoppedRunFinally`

src/hacbs/components/PipelineRunListView/pipelinerun-actions.tsx
mfrances17: changed it so it is only enabled when the PLR is running, thanks

src/hacbs/types/pipelineRun.ts
mfrances17: done, thanks!

christianvogt:
/approve

Talk with Karthik as it seems for prekcp we are running an old version of pipelines and the status I told you to use doesn't work here :(

PR #207, opened 2022-08-30
fix useSearchParam using stale value with multiple update

christianvogt:
## Description
If there are multiple uses of `useSearchParam` where the search params are updated multiple times in the same thread of execution, after the first update, subsequent updates are using a stale value of the previous search params.

The fix is to read directly from `window.location.search` when performing an update .

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

## How to test or reproduce?

- Enable HACBS.
- Go to `Settings` page
- Ensure there is at least one environment; create if necessary
- Enter a name filter that matches no environments
- Set a type filter
- When there are no matches, the empty state shows a "Clear all filters" button
- Click "clear all filters"

Observe that it requires 2 clicks to clear all the filters because the first click only clears the type filter. 

With the fix, both the type filter and name filter would be cleared.

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

cc @jeff-phillips-18

### DIFF
--- a/src/hacbs/components/Environment/__tests__/EnvironmentListView.spec.tsx
+++ b/src/hacbs/components/Environment/__tests__/EnvironmentListView.spec.tsx
@@ -3,8 +3,11 @@ import '@testing-library/jest-dom';
 import { useK8sWatchResource } from '@openshift/dynamic-plugin-sdk-utils';
 import { render, screen, configure, cleanup, fireEvent } from '@testing-library/react';
 import { EnvironmentKind } from '../../../../types';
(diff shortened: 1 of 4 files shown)


jeff-phillips-18 (approved):
Tested locally. Works great.

/lgtm

christianvogt:
@jeff-phillips-18 I pushed an update because the `EnvironmentListView.spec.tsx` test case was failing without mocking `window.location.search`

PR #208, opened 2022-08-30
Replace instances of ReleaseLink with ReleasePlan

ralphbean:
## Description

The API changed in https://github.com/redhat-appstudio/infra-deployments/pull/648. I don't know how I didn't see this earlier in the day, but for me now the overview tab fails to render, making repeated requests for ReleaseLinks which no longer exist (renamed to ReleasePlans).

## Type of change

- [x] Bugfix

## How to test or reproduce?

I did not test this change locally - but, you should be able to verify if you can view the overview tab with `?hacbs=true` pointed at the current app studio stage cluster.

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


src/hacbs/types/coreBuildService.ts
```diff
@@ -51,13 +51,18 @@ export type Env = {
...
+export type ReleasePlanKind = K8sResourceCommon & {
+  spec: ReleasePlanSpec;
 };
 
-export type ReleaseLinkSpec = {
+export type ReleasePlanSpec = {
   displayName: string;
   application: string;
-  target: string;
+  target: ReleaseTarget;
```
scoheb, line 61: should this be `ReleaseTarget.namespace` for now?

ralphbean: I don't think so - I think this defines a new type, and this line indicates that the `target` of the `ReleasePlanSpec` should be another typed object. That type is defined just below this one, and it contains a mandatory `namespace` and an optional `workspace`.

PR #221, opened 2022-09-21
Commit-status-and-panel

abhinandan13jan:
## Fixes 

https://issues.redhat.com/browse/HACBS-1136

## Description

Adds status and task side panel to commit details

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

Screenshot 2022-10-10 at 6 33 39 PM

Screenshot 2022-10-10 at 6 33 30 PM

Screenshot 2022-10-14 at 7 27 47 PM
Screenshot 2022-10-14 at 7 28 10 PM

## How to test or reproduce?

Added unit tests for commit-utils, CommitSidePanelHeader, CommitSidePanelbody

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/hacbs/components/Commits/CommitDetailsView.tsx
+++ b/src/hacbs/components/Commits/CommitDetailsView.tsx
@@ -2,17 +2,19 @@ import React from 'react';
 import { useK8sWatchResource } from '@openshift/dynamic-plugin-sdk-utils';
 import {
   Bullseye,
+  Button,
+  Drawer,
+  DrawerContent,
+  DrawerContentBody,
   EmptyState,
   EmptyStateBody,
   EmptyStateIcon,
-  Flex,
-  FlexItem,
   Spinner,
   Text,
+  TextVariants,
   Title,
 } from '@patternfly/react-core';
 import { ExclamationCircleIcon } from '@patternfly/react-icons/dist/esm/icons/exclamation-circle-icon';
-import { ExternalLinkAltIcon } from '@patternfly/react-icons/dist/esm/icons/external-link-alt-icon';
 import { SearchIcon } from '@patternfly/react-icons/dist/js/icons';
 import { GithubIcon } from '@patternfly/react-icons/dist/js/icons/github-icon';
 import { pipelineRunFilterReducer } from '../../../shared';
@@ -23,9 +25,11 @@ import { useNamespace } from '../../../utils/namespace-context-utils';
 import { PipelineRunLabel } from '../../consts/pipelinerun';
 import { PipelineRunGroupVersionKind } from '../../models/';
 import { PipelineRunKind } from '../../types';
-import { createCommitObjectFromPLR, getCommitShortName } from '../../utils/commits-utils';
+import { createCommitObjectFromPLR, getCommitShortName, statuses } from '../../utils/commits-utils';
 import DetailsPage from '../ApplicationDetails/DetailsPage';
 import CommitsGettingStartedModal from './CommitsGettingStartedModal';
+import CommitSidePanel from './CommitSidePanel';
+import { SortedPLRList } from './CommitSidePanelHeader';
 import CommitsOverviewTab from './tabs/CommitsOverviewTab';
 import CommitsPipelineRunTab from './tabs/CommitsPipelineRunTab';
 
@@ -47,6 +51,23 @@ const CommitDetailsView: React.FC<CommitDetailsViewProps> = ({ commitName, appli
       window.localStorage.setItem(COMMITS_GS_LOCAL_STORAGE_KEY, 'true');
     }
     setShowGettingStarted(shown);
+  };
+
+  const [isExpanded, setIsExpanded] = React.useState<boolean>(false);
+  const [isFull, setIsFull] = React.useState<boolean>(false);
+  const [selectedPipelineRun, setSelectedPipelineRun] = React.useState<PipelineRunKind>(null);
+  const drawerRef = React.useRef<HTMLDivElement>();
+
+  // const onPanelExpand = () => {
+  //   drawerRef.current && drawerRef.current.focus();
+  // };
+
+  const onStatusClick = () => {
+    setIsExpanded(!isExpanded);
+  };
+
+  const onPanelCloseClick = () => {
+    setIsExpanded(false);
   };
 
   const [pipelineruns, loaded, loadErr] = useK8sWatchResource<PipelineRunKind[]>({
@@ -71,19 +92,47 @@ const CommitDetailsView: React.FC<CommitDetailsViewProps> = ({ commitName, appli
     [loaded, pipelineruns],
   );
 
-  const status = React.useMemo(
-    () =>
-      loaded &&
-      Array.isArray(pipelineruns) &&
-      pipelineruns.length > 0 &&
-      pipelineRunFilterReducer(
-        pipelineruns.sort(
-          (a, b) =>
-            parseInt(a.metadata.creationTimestamp, 10) - parseInt(b.metadata.creationTimestamp, 10),
-        )[pipelineruns.length - 1],
-      ),
-    [loaded, pipelineruns],
-  );
+  const sortedPLRList: SortedPLRList = React.useMemo(() => {
+    if (!loaded || loadErr) {
+      return null;
+    }
+
+    const runs = {};
+    pipelineruns.forEach((plr) => {
+      // sort plr into respective status array
+      const plrStatus = pipelineRunFilterReducer(plr);
+      // put task in correct status array
+      statuses.forEach((status) => {
+        if (plrStatus === status) {
+          if (!runs[status]) {
+            runs[status] = [];
+          }
+          runs[status].push(plr);
+        }
+      });
+    });
+
+    // sort each status array
+    statuses.forEach((status) => {
+      if (runs[status] && Array.isArray(runs[status])) {
+        runs[status].sort((a, b) => parseInt(a.startTime, 10) - parseInt(b.startTime, 10));
+      }
+    });
+
+    return runs;
+  }, [pipelineruns, loaded, loadErr]);
+
+  const commitStatus = React.useMemo(() => {
+    if (sortedPLRList) {
+      for (const s in sortedPLRList) {
+        if (sortedPLRList[s]?.length > 0) {
+          setSelectedPipelineRun(sortedPLRList[s][0]);
+          return s;
+        }
+      }
+    }
+    return '-';
+  }, [sortedPLRList]);
 
   const commitDisplayName = getCommitShortName(commitName);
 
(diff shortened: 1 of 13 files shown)


Ranelim:
@abhinandan13jan looks very cool so far!

Few comments:

- Let's move the GitHub icon and the status label after the page title. I think @christianvogt mentioned it is doable now
- Did we include a "read more" after a long commit comment? I think we set up a max character of 72 and "read more" will just display the rest.
- On the side panel:
- I'm not sure what is "go-3-on-push..." is that a build pipeline placeholder?
- We should have a side panel title, displaying the full name of the node
- Below that, we should have an icon and generic name of the node type (in grey, body weight)
- For the status below, we should start that string with the corresponding icon (green/red/yellow/blue)

Thank you!

abhinandan13jan:
> @abhinandan13jan looks very cool so far!
> 
> Few comments:
> 
> * Let's move the GitHub icon and the status label after the page title. I think @christianvogt mentioned it is doable now
> * Did we include a "read more" after a long commit comment? I think we set up a max character of 72 and "read more" will just display the rest.
> * On the side panel:
> * I'm not sure what is "go-3-on-push..." is that a build pipeline placeholder?
> * We should have a side panel title, displaying the full name of the node
> * Below that, we should have an icon and generic name of the node type (in grey, body weight)
> * For the status below, we should start that string with the corresponding icon (green/red/yellow/blue)
> 
> Thank you!

1. Moved the GH icon to header
2. Yes, I updated to the read more, but this string is only 56 char long so it doesn't show. However, for demo I've set the read more length to 45 in screenshot
3. Go-random-number is the actual kubernetes object name for the pipeline. I've updated to use logical name "build/Release/Integration Test" , `Buld` in this case instead. We currently don't have the logical mapping/definitions in place with the `real object`(kubernetes PipelineRun CR) vs the `display texts`, which will be taken care of once @karthikjeeyar data model for the workflow visualization is in

src/hacbs/components/Commits/CommitDetailsView.tsx
```diff
@@ -106,112 +157,151 @@ const CommitDetailsView: React.FC<CommitDetailsViewProps> = ({ commitName, appli
...
+                  <span className="pf-u-mr-sm">
+                    <b>{commitDisplayName}</b>
+                  </span>
+                  <Button
+                    className="pf-u-pl-xs"
+                    aria-expanded={isExpanded}
+                    variant="plain"
+                    onClick={onStatusClick}
+                  >
+                    <StatusIconWithText status={commitStatus} />
```
karthikjeeyar, line 210: This component is rendering [PF badge](https://patternfly-react.surge.sh/components/badge) component, but the designs has[ PF Label ](https://patternfly-react.surge.sh/components/label)component with background colors matching the status.

current: 
image

designs:
image

src/hacbs/components/Commits/CommitSidePanel.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+const CommitSidePanel: React.FC<CommitSidePanelProps> = ({
+  drawerRef,
+  isExpanded,
+  onPanelCloseClick,
+  initialStatus,
+  setSelectedPipelineRun,
+  sortedPLRList,
+}) => {
+  const [currentStatus, setCurrentStatus] = React.useState<string>(initialStatus);
```
karthikjeeyar, line 48: In this case, If you are expecting to sync your props (initialStatus) with useState variable then this will fail to sync. UseState will not re-initialise when the props change. You need to sync it either with useEffect or re-render this component with `key`.

karthikjeeyar: Also the fact that this status will change on progression of events, better to call something other than `initialStatus`, maybe `currentStatus` ?

karthikjeeyar (commented):
Screenshot 2022-09-27 at 12 41 53 PM

Choosing a different status in dropdown breaks the UI.

src/hacbs/components/Commits/CommitSidePanel.tsx
abhinandan13jan: Renamed to commitStatus and moved inside useEffect

abhinandan13jan:
Updated PR
- Modularised CommitSidePanel into Two new components CommitSidePanelHeader and CommitSidePanelBody
- Updated `StatusIconWithText` to use `Label` instead of `Badge`
- Added test suites for CommitSidePanelHeader, CommitSidePanelBody & commit-utils
- Moved initialStatus as commitStatus inside useEffect

abhinandan13jan:
Screenshot 2022-09-29 at 11 53 05 PM

src/shared/components/pipeline-run-logs/StatusIcon.tsx
```diff
@@ -66,17 +66,26 @@ export const StatusIconWithText: React.FC<StatusIconProps & { text?: string }> =
   ...others
 }) => {
   return (
-    <Badge isRead>
+    <Label>
```
karthikjeeyar, line 69: Label here is missing color property, so the background is still grey.

image

src/shared/components/pipeline-run-logs/StatusIcon.tsx
```diff
@@ -66,17 +66,26 @@ export const StatusIconWithText: React.FC<StatusIconProps & { text?: string }> =
...
-      {text ?? status}
-    </Badge>
+      </span>
+      <span
+        style={{
+          color: status
+            ? getRunStatusColor(status).pftoken.value
+            : getRunStatusColor(runStatus.Cancelled).pftoken.value,
+        }}
+      >
```
karthikjeeyar, line 86: Refer my previous comment.  Once color prop is added on the label, you no longer need to set the color on icon, text seperately.

abhinandan13jan: updated
Screenshot 2022-10-10 at 6 33 30 PM

src/hacbs/components/Commits/CommitSidePanelHeader.tsx
```diff
@@ -0,0 +1,118 @@
...
+        </Button>
+        <Button
+          variant={ButtonVariant.control}
+          className="pf-u-p-sm"
+          isDisabled={currentIndex < 1}
+          onClick={previousTask}
+          data-test="previous-plr-button"
+        >
+          <AngleUpIcon />
+        </Button>
```
christianvogt, line 109: The spacing on these buttons don't look like the UX doc.
This is something we can also address afterwards.

But it seems like UX wants what looks like input group attachments. I'm not sure exactly though.
@Ranelim could you help identify the PF components you've used in this mock:
image

src/hacbs/utils/commits-utils.ts
```diff
@@ -77,3 +86,38 @@ export const getCommitsFromPLRs = (plrList: PipelineRunKind[], limit?: number):
...
 
 export const getCommitShortName = (commitName: string): string => commitName.slice(0, 7);
+
+export const showPLRType = (plr: PipelineRunKind): string => {
+  if (!plr) {
+    return '-';
+  }
+  const runType = plr?.metadata.labels[PipelineRunLabel.COMMIT_TYPE_LABEL];
+  if (!runType) {
+    return plr.metadata.name;
```
christianvogt, line 96: This seems like an incorrect value for when there is no type available.

src/hacbs/components/Commits/CommitSidePanel.tsx
```diff
@@ -0,0 +1,87 @@
...
+  commitStatus,
+  setSelectedPipelineRun,
+  sortedPLRList,
+}) => {
+  const [currentStatus, setCurrentStatus] = React.useState<string>(commitStatus);
+  const [currentIndex, setCurrentIndex] = React.useState<number>(0);
+
+  React.useEffect(() => {
+    setCurrentStatus(commitStatus);
+  }, [commitStatus]);
```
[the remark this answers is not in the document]
abhinandan13jan: For a pipeline in progress, the commitstatus would change. So we not capture that?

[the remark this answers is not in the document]
christianvogt: But doesn't this cause the view to switch on the user while they may be viewing a different one?

src/hacbs/components/Commits/CommitDetailsView.tsx
```diff
@@ -71,19 +92,51 @@ const CommitDetailsView: React.FC<CommitDetailsViewProps> = ({ commitName, appli
...
+
+    pipelineruns.map((plr) => {
+      // sort plr into respective status array
+      const plrStatus = pipelineRunFilterReducer(plr);
+      // put task in correct status array
+      statuses.map((status) => {
+        if (plrStatus === status) {
+          runs[status].push(plr);
+        }
+      });
```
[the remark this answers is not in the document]
abhinandan13jan: No, the `statuses` array holds the order specified by UX for different `runStatuses`

src/hacbs/components/Commits/CommitSidePanel.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  const changeStatus = (status) => {
+    status !== currentStatus && setCurrentStatus(status);
+    currentIndex !== 0 && setCurrentIndex(0);
```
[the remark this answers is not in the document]
abhinandan13jan: When user clicks on nextTask or PreviousTask, current index changes

src/hacbs/components/Commits/CommitDetailsView.tsx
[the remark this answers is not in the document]
christianvogt: Sorry I corrected my suggestion. Not sure why i wrote `runStatus` while i meant `run`. 
See now.

The suggestion is to simply look at the current `run` status map and if it contains a value, we push to that array rather than having to perform a loop to find a match with `statuses` since `run` should have already be initialized to only contain entries that match those in `statuses`.

abhinandan13jan:
updated @christianvogt @karthikjeeyar 
Screenshot 2022-10-14 at 7 27 47 PM
Screenshot 2022-10-14 at 7 28 10 PM

src/hacbs/components/Commits/CommitDetailsView.tsx
```diff
@@ -49,6 +53,23 @@ const CommitDetailsView: React.FC<CommitDetailsViewProps> = ({ commitName, appli
...
   };
 
+  const [isExpanded, setIsExpanded] = React.useState<boolean>(false);
+  const [isFull, setIsFull] = React.useState<boolean>(false);
+  const [selectedPipelineRun, setSelectedPipelineRun] = React.useState<PipelineRunKind>(null);
+  const drawerRef = React.useRef<HTMLDivElement>();
+
+  // const onPanelExpand = () => {
+  //   drawerRef.current && drawerRef.current.focus();
+  // };
```
karthikjeeyar, line 63: nit: remove commented code

src/hacbs/components/Commits/CommitDetailsView.tsx
```diff
@@ -71,19 +92,47 @@ const CommitDetailsView: React.FC<CommitDetailsViewProps> = ({ commitName, appli
...
+      // put task in correct status array
+      statuses.forEach((status) => {
+        if (plrStatus === status) {
+          if (!runs[status]) {
+            runs[status] = [];
+          }
+          runs[status].push(plr);
+        }
+      });
+    });
```
karthikjeeyar, line 113: you missed to change the suggestion from christian to remove the status map and directly push the `plrStatus` in runs array.

karthikjeeyar (commented):
Works as expected, still one more comment from christian needs to be resolved.

src/hacbs/components/Commits/CommitSidePanelHeader.tsx
christianvogt: icon buttons don't have grey outlines. This is what confused me. If we don't need to have those grey outlines then we're good. I thought it was some other component / styling i wasn't aware of.

PR #222, opened 2022-09-21
View components & environment flow while viewing details

jeff-phillips-18:
## Fixes 
Fixes https://issues.redhat.com/browse/HAC-2151

## Description
Updates the application details view to use the environment cards as tabs to show the application environment details view below on the same page.

Updates PF versions to pick fix for topology view sizing

## Type of change

- [x] Feature

image

image

image

image

image

/cc @itsptk @serenamarie125 @christianvogt @karthikjeeyar


src/components/Environment/ApplicationEnvironmentCards.tsx
```diff
@@ -74,58 +89,43 @@ const ApplicationEnvironmentCard: React.FC<ApplicationEnvironmentCardProps> = ({
...
+              {isPositionedEnvironment(env, environments.slice(0, index)) ? (
+                <div className="application-environment-cards__env-arrow">
+                  <ArrowRightIcon />
+                </div>
+              ) : null}
+              <ApplicationEnvironmentCard
+                environment={env}
+                isExpanded={cardsExpanded}
+                isSelected={env.metadata.name === selectedEnvironment}
+                onSelect={() => onSelect && onSelect(env.metadata.name)}
```
karthikjeeyar, line 119: we can use same displayName in the heading below like we have in card title.

image


```suggestion
                onSelect={() => onSelect && onSelect(env.spec.displayName)}
```

karthikjeeyar:
When there is getting started card on the top of the page, then the graph area is too small to work with (height 150px). 
 
image

src/components/Environment/ApplicationEnvironmentCards.tsx
rohitkrai03: Agreed, I think it would be better to use display name here as well.

rohitkrai03 (approved):
/approve

rohitkrai03:
@jeff-phillips-18 Can you update the PR with the small change that @karthikjeeyar suggested? We can merge the PR after that.

jeff-phillips-18:
@rohitkrai03 Updated and squashed. Thanks!

jeff-phillips-18:
@karthikjeeyar I've reverted your requested change and updated only where we display the title to use the `spec.displayName`. Using the displayName for determining selection could be problematic since it is not guaranteed to be unique.

PR #238, opened 2022-10-14
HACBS UI test: Pipelineruns

dheerajodha:
## Description

- This PR adds another HACBS-related UI test, which tests the triggering of pipelineruns.
- This is done by using 2 yaml files, run via a script called `trigger-pipelineruns-script.sh` which `oc apply`ies both these files and hence triggers pipelineruns.
- The test then ensures that the pipelineruns are completed successfully by checking their status on various pages on the UI.

[Last successful test recording](https://drive.google.com/file/d/1QC1vN0xpcRXHyC2UIfmvwGGl3r0B2erK/view?usp=sharing)

NOTE:
- The test was working until yesterday but started failing due to some issue with the pipeline service on kcp-stable. So, I'd not get this merged until this test works locally again. I'll keep trying in the meantime, but this PR is open for review.
- I'd need to update the test after the [PR#204](https://github.com/openshift/hac-dev/pull/204) gets merged because it introduces some UI changes to the pipelinerun details page.

## Type of change

- [x] Other (please describe): Test

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


dheerajodha:
The test is failing because the current cypress image from quay.io doesn't have `which` (and `oc`) packages installed.
Waiting for this PR (https://github.com/openshift/hac-dev/pull/247) to get merged, which introduces a custom Dockerfile that builds on top of the above-mentioned cypress image and installs the required packages in subsequent layers.

PR #242, opened 2022-10-20
Fix extraspace source

abhinandan13jan:
## Fixes 

https://issues.redhat.com/browse/HACBS-1203

## Description

Only trims during validation and not the actual value

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

Screenshot 2022-10-19 at 7 25 46 PM
Screenshot 2022-10-19 at 7 25 52 PM

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


src/components/ImportForm/utils/auth-utils.ts
```diff
@@ -24,6 +24,7 @@ import { useNamespace } from '../../../utils/namespace-context-utils';
 export const useAccessCheck = (
   source: string,
   dependency,
+  needToTrim?,
```
christianvogt, line 27: Either we always trim inside this utility function, or we rely on the caller to have trimmed the value. But it's unnecessary to parameterize this utility function with the ability to trim or not trim.

src/components/ImportForm/SourceSection/SourceSection.tsx
```diff
@@ -55,7 +55,7 @@ export const SourceSection: React.FC<SourceSectionProps> = ({
   const label = gitOnly ? 'Git repo URL' : 'Git repo URL or container image';
 
   const [{ isGit, isRepoAccessible, serviceProvider, accessibility }, accessCheckLoaded] =
-    useAccessCheck(sourceUrl, authSecret);
+    useAccessCheck(sourceUrl, authSecret, true);
```
christianvogt, line 58: Either trim the sourceUrl here or inside the utility function for every use case.

src/components/ImportForm/utils/auth-utils.ts
abhinandan13jan: updated

src/components/ImportForm/SourceSection/SourceSection.tsx
abhinandan13jan: updated

PR #243, opened 2022-10-20
Remove tokens and bindings after GH private test.

Katka92:
## Description
Remove tokens and token bindings after GH private repo test to enable testing another way of a GH Sign up. This new test will be delivered in another separate PR.


integration-tests/utils/Tokens.ts
```diff
@@ -0,0 +1,40 @@
...
+            removeAllResources(token, namespace, "spiaccesstokenbinding");
+            removeAllResources(token, namespace, "spiaccesstoken");
+        });
+    } 
+}
+
+function removeAllResources(token: string, namespace: string, resourceType : string) {
+    const getResources = {
+        method: 'GET',
+        url: `https://env-ephemeral-alqcwz-yjsfmfej.apps.c-rh-c-eph.8p0c.p1.openshiftapps.com/api/k8s/apis/appstudio.redhat.com/v1beta1/namespaces/${namespace}/${resourceType}s`,
```
jrichter1, line 17: the only reason this url works is because there is a long lived ephemeral env behind it

this needs to be parameterized

integration-tests/utils/Tokens.ts
```diff
@@ -0,0 +1,40 @@
...
+            accept: 'application/json',
+        }
+    }
+
+    cy.request(getResources).its('body').then(respBody => {
+        for (const item of respBody.items) {
+            let resourceName = item.metadata.name;
+            const removeResources = {
+                method: 'DELETE',
+                url: `https://env-ephemeral-alqcwz-yjsfmfej.apps.c-rh-c-eph.8p0c.p1.openshiftapps.com/api/k8s/apis/appstudio.redhat.com/v1beta1/namespaces/${namespace}/${resourceType}s/${resourceName}`,
```
jrichter1, line 29: same as above, might be better to extract the url to a variable so it's guaranteed consistency between the requests

jrichter1 (changes requested):

PR #247, opened 2022-11-01
Add our own dockerfile

Katka92:
## Description
Add our own Dockerfile and run tests with our own image. 
Now, the image has also `oc` and `which` installed.
The image is updated and pushed manually.

### DIFF
--- a/integration-tests/docker/Dockerfile
+++ b/integration-tests/docker/Dockerfile
@@ -7,4 +7,6 @@ ENV OC_DOWNLOAD_URL="https://downloads-openshift-console.apps.hac-devsandbox.5un
 RUN microdnf install --assumeyes which && \
     curl -L -o /tmp/oc.tar ${OC_DOWNLOAD_URL} && \
     tar -xf /tmp/oc.tar -C /usr/bin
-    
+
+RUN groupadd -r hacdev && useradd -r -g hacdev hacdev
+USER hacdev:hacdev
--- a/integration-tests/docker/Dockerfile
+++ b/integration-tests/docker/Dockerfile
@@ -1,0 +1,10 @@
+FROM quay.io/redhatqe/cypress:9.6.0
+
+USER root:root
+
+ENV OC_DOWNLOAD_URL="https://downloads-openshift-console.apps.hac-devsandbox.5unc.p1.openshiftapps.com/amd64/linux/oc.tar"
+
+RUN microdnf install --assumeyes which && \
+    curl -L -o /tmp/oc.tar ${OC_DOWNLOAD_URL} && \
+    tar -xf /tmp/oc.tar -C /usr/bin
+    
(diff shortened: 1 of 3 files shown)


dheerajodha:
Thank you @Katka92 for this PR! I can confirm that now I'm able to use `oc` and `which` using this image.

> (python) [djodha@djodha docker]$ docker run -it --entrypoint=/bin/bash quay.io/hacdev/hac-tests
bash-5.1# oc version
Client Version: 4.10.0-202210061437.p0.g442535c.assembly.stream-442535c
bash-5.1# 
bash-5.1# which
Usage: which [options] [--] COMMAND [...]

integration-tests/docker/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM quay.io/redhatqe/cypress:9.6.0
+
+USER root:root
```
dheerajodha, line 3: I'm not a docker expert but using a root user is not a good idea ([see this](https://security.stackexchange.com/questions/106860/can-a-root-user-inside-a-docker-lxc-break-the-security-of-the-whole-system)).
Maybe we can use a non-root user ([see this](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user)) by adding following commands in place of line 3:
`RUN groupadd -r hacdev && useradd -r -g hacdev hacdev`
`USER hacdev:hacdev`

Katka92: I'm neither a docker expert :wink: But from what I see, we need root permissions to install a new software in this container. When I put  your commands instead the `USER root:root` it fails on:
```
groupadd: Permission denied.
```
If I do that like a root and then switch to a new user, I'm getting an error during installation of `which` :
```
error: Failed to create: /var/cache/yum/metadata
```
So I guess that way the `hacdev` user doesn't have enough permissions. 
Anyway, this root user is used for **building** the container, so I would say it may be OK to leave the root here.

dheerajodha: Sorry for not testing locally before adding my previous comment.
One way is to install all the relevant packages as a root user and then switch to a non-user after that ([I referred this](https://stackoverflow.com/a/54269736)). I tested locally by adding 
> RUN groupadd -r hacdev && useradd -r -g hacdev hacdev
USER hacdev:hacdev

to the end of the Dockerfile, the image gets built successfully with the root staying in its original place. I'm also curious about what @jrichter1 thinks about this.

jrichter1: AFAIK the test commands are called with `-u 0`, which probably is root anyway. Would be nice to have a non root user, maybe it would eliminate the need to `chmod a+rwx` for the build artifacts.

jrichter1: let root do the installation, you need the permissions for that, but use a non root to run the tests

rohitkrai03 (approved):
/approve

PR #249, opened 2022-11-03
add optional label to mark a integration test as optional

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/HACBS-1362

## Description

Add `test.appstudio.openshift.io/optional: true/false` to mark an `integrationTestScenario` optional instead of `spec.optional` (no longer valid).

## Type of change

- [x] Bugfix

## How to test or reproduce?

1. Add an Integration test and mark it as optional.
2. The integration test CR should contain the optional labels.

## Unit tests
```
  Create Utils
    ✓ Should contain the optional test value in the labels (1 ms)
    ✓ Should not contain the optional label if the test is mandatory
```

```
  IntegrationTestListRow
    ✓ should read the optional field from label (8 ms)
    ✓ should show mandatory value if the optional labels are not set (8 ms)
```

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

### DIFF
--- a/src/hacbs/components/ImportForm/create-utils.ts
+++ b/src/hacbs/components/ImportForm/create-utils.ts
@@ -6,11 +6,12 @@ import {
   IntegrationTestScenarioModel,
 } from '../../models/hacbs';
 import { IntegrationTestScenarioKind } from '../../types/coreBuildService';
-import { FormValues, IntegrationTestFormValues } from './types';
-
-export enum IntegrationTestAnnotations {
-  DISPLAY_NAME = 'app.kubernetes.io/display-name',
-}
+import {
+  FormValues,
+  IntegrationTestAnnotations,
+  IntegrationTestFormValues,
+  IntegrationTestLabels,
+} from './types';
 
 // /**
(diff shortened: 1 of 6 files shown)


karthikjeeyar:
FYI @mfrances17 I see you are working on edit integration test,  with this change we need to read the value from the label and not from `spec`.

cc:  @christianvogt

christianvogt:
@karthikjeeyar please update `IntegrationTestListRow` to read the optional property from the labels.

src/hacbs/components/ImportForm/create-utils.ts
```diff
@@ -36,6 +37,7 @@ export const createIntegrationTest = (
     metadata: {
       name,
       namespace,
+      ...(optional && { labels: { [IntegrationTestLabels.OPTIONAL]: optional.toString() } }),
```
abhinandan13jan, line 40: just out of curiosity...is it the only label an integrationTestScenario would carry?

PR #253, opened 2022-11-04
Send context dir while creating components

rohitkrai03:
## Fixes 

- https://issues.redhat.com/browse/HAC-2376

## Description

- Due to an earlier HAS issue we could not send the context dir value while creating components. That issue has been fixed so this PR updates the code to send the context value.

## Type of change

- [x] Bugfix

- No UI change.

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/utils/create-utils.ts
+++ b/src/utils/create-utils.ts
@@ -85,11 +85,6 @@ export const createComponent = (
   const { componentName, containerImage, source, replicas, resources, env, targetPort } = component;
 
   const name = component.componentName.split(/ |\./).join('-').toLowerCase();
-  // const uniqueName = uniqueId(name);
-
-  // FIXME - context is not supported by the HAS API correctly yet.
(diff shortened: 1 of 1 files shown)


sahil143:
@rohitkrai03 correct context is not sent with create component api call

- With CDQ call
image

- with create component call
image

rohitkrai03:
> @rohitkrai03 correct context is not sent with create component api call
> 
> * With CDQ call
> 
> image
> 
> * with create component call
> 
> image

@sahil143 We are actually sending correct context with the CDQ call as well as component call. The difference is we are sending the result of CDQ to create component call and some times the CDQ returns different context than what was sent to it. This is an issue on the HAS side where they do not consider context while detecting and its been fixed on KCP environment but not on devsandbox. So our code should be good to go.

sahil143 (approved):

PR #260, opened 2022-11-22
Fix component creation without targetPort

karthikjeeyar:
## Fixes 

[HACBS-1008](https://issues.redhat.com/browse/HACBS-1008)

## Description

When trying to create a component that doesn't need to use any port, if I leave the "Target port" text box empty, nothing happens when I click on the "Create components & continue".

## Type of change

- [x] Bugfix

**Before fix:** 

without_targetPort_before

**After fix:** 
without_targetPort_after

## How to test or reproduce?

 In component review section, remove the tagertPort value under `Deploy Configuration` section and click on `create component & continue` button

## Unit tests:
```
Create Utils
    ✓ Should create component with target port
    ✓ Should create component without target port, if it is not passed (1 ms)
```
## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

### DIFF
--- a/src/utils/__tests__/create-utils.spec.ts
+++ b/src/utils/__tests__/create-utils.spec.ts
@@ -170,6 +170,46 @@ describe('Create Utils', () => {
     });
   });
 
+  it('Should create component with target port', async () => {
+    const mockComponentDataWithTargetPort = {
+      ...mockComponent,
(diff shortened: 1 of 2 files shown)


karthikjeeyar:
cc: @christianvogt

PR #271, opened 2022-11-30
edit integration test scenario

christianvogt:
## Fixes 
https://issues.redhat.com/browse/HACBS-1172

## Description

Adds support to edit integration test scenarios.
Enables the `Edit` action on the details page.

Updated copy for Add integration test scenario form which is used in both the wizard and stand alone add page. 

The `Integration test name` input is marked as read only and cannot be edited.
The `Save changes` button only gets enabled if the edit form has changes.

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

image

empty state if the URL points to an invalid integration test scenario name:
image

Add integration test scenario:
image

Add integration test scenario in wizard:
image

Add integration test scenario standalone page:
image

Invalid name error:
image

## How to test or reproduce?
Add integration test scenario through wizard.
Add integration test scenario from application details page action menu.
Edit integration test scenario from its details page action menu.

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/config/remotePlugin.js
+++ b/config/remotePlugin.js
@@ -6,6 +6,7 @@ const hacbs = {
     PipelineRuns: resolve(__dirname, '../src/hacbs/pages/PipelineRunPage'),
     HACBSFlag: resolve(__dirname, '../src/hacbs/hacbsFeatureFlag'),
     HACBSImport: resolve(__dirname, '../src/hacbs/pages/ImportPage'),
+    EditIntegrationTest: resolve(__dirname, '../src/hacbs/pages/EditIntegrationTestPage'),
     HACBSIntegrationTest: resolve(__dirname, '../src/hacbs/pages/IntegrationTestPage'),
     CommitsPage: resolve(__dirname, '../src/hacbs/pages/CommitsPage'),
     IntegrationTestDetails: resolve(__dirname, '../src/hacbs/pages/IntegrationTestDetailsPage'),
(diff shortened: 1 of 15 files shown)


MariaLeonova:
@christianvogt could we move the error message closer to the related field as [per guidelines here](https://www.patternfly.org/v4/components/alert/design-guidelines/#using-bordered-inline-alerts)? I'm looking at the [last screenshot](https://user-images.githubusercontent.com/14068621/204925081-33b09354-c8f4-401c-a42e-bec62b90bb99.png).

@abigaeljamie and I discussed the error message as well, when we were working on naming issues. And if at all possible we'd love to have it updated to [this suggestion](https://docs.google.com/document/d/1FxtJb72o3kaUk0m-boO2pG9GMg_2S7u-C1h6eBEfA54/edit#bookmark=id.b076m3tee5w2).

christianvogt:
@MariaLeonova this error is coming from the server and not a client side validation. I don't believe there's enough structured data to know which field has caused the error. Yes there is `metadata.name` in the error message description itself, but that's not structured enough to rely upon in general.

As we merge with app studio, I'll work with Rohit to see if there's a consistent way we can associate errors with related fields. However I feel like we've gone down this route already, not just hacbs, and concluded that we can't make the association.

MariaLeonova:
@christianvogt understood! Can we at least make the message more human-readable?

christianvogt:
@MariaLeonova Sorry I did update the code to use the better error message. Let me get an updated screenshot.

christianvogt:
@MariaLeonova I've updated the screenshot just now.

PR #281, opened 2022-12-12
Refactor hacbs folder and merge into src

jeff-phillips-18:
## Fixes 
Fixes [HAC-2493](https://issues.redhat.com/browse/HAC-2493)

## Description
Move HACBS code up into src directory. Remove hacbs specificity on local storage items. Remove hacbs css prefix.

A few components were left in the `hacbs` directory due to naming conflicts. These components need to get resolved to be the same component regardless of the feature flag (that is, the component should look the same), waiting on UX to determine the common look of these components.

## Type of change
- [x] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no api changes)


src/components/Modals/GettingStartedModal.tsx
```diff
@@ -11,10 +11,10 @@ import {
   AlertActionLink,
```
rottencandy, line 11: We already have `src/components/modal`, should we move modals there instead of having two `modal` and `Modals` directories?

src/components/Components/__tests__/ComponentListView.spec.tsx
```diff
@@ -2,12 +2,12 @@ import * as React from 'react';
 import '@testing-library/jest-dom';
 import { useK8sWatchResource } from '@openshift/dynamic-plugin-sdk-utils';
 import { render, screen, fireEvent, within } from '@testing-library/react';
-import { WatchK8sResource } from '../../../../dynamic-plugin-sdk';
-import { mockRoutes } from '../../../../hooks/__data__/mock-data';
-import { useApplicationRoutes } from '../../../../hooks/useApplicationRoutes';
-import { useGitOpsDeploymentCR } from '../../../../hooks/useGitOpsDeploymentCR';
-import { ComponentGroupVersionKind } from '../../../../models';
-import { PipelineRunGroupVersionKind } from '../../../../shared';
+import { WatchK8sResource } from '../../../dynamic-plugin-sdk';
```
rottencandy, line 5: Not important, but I think we no longer need the `dynamic-plugin-sdk.ts` file. All sdk types are now available in `@openshift/dynamic-plugin-sdk-utils`.

src/EnterpriseContractView/EnterpriseContractView.tsx
```diff
@@ -15,14 +15,11 @@ import {
...
 import { ExternalLinkAltIcon } from '@patternfly/react-icons/dist/esm/icons/external-link-alt-icon';
-import ExternalLink from '../../../shared/components/links/ExternalLink';
-import { useNamespace } from '../../../utils/namespace-context-utils';
-import { EnterpriseContractPolicyGroupVersionKind } from '../../models';
-import { EnterpriseContractPolicyKind } from '../../types/enterpriseContractPolicy';
-import {
-  HACBS_ENTERPRISE_CONTRACT_INFO_LINK,
-  HACBS_ENTERPRISE_CONTRACT_POLICIES_DATA,
-} from './const';
+import { EnterpriseContractPolicyGroupVersionKind } from '../models';
```
rohitkrai03, line 18: Shouldn't these components also go inside `src/components`?

src/components/IntegrationTest/IntegrationTestDetailsView.tsx
```diff
@@ -13,12 +13,12 @@ import {
 } from '@patternfly/react-core';
 import { ExclamationCircleIcon } from '@patternfly/react-icons/dist/esm/icons/exclamation-circle-icon';
 import { SearchIcon } from '@patternfly/react-icons/dist/js/icons';
-import { useModalLauncher } from '../../../components/modal/ModalProvider';
-import { useNamespace } from '../../../utils/namespace-context-utils';
 import { IntegrationTestScenarioGroupVersionKind } from '../../models';
```
rohitkrai03, line 16: I see `IntegrationTest`, `IntegrationTestForm` and `IntegrationTestListView`, maybe all the integration tests related components can go into one IntegrationTest folder?

rohitkrai03 (changes requested):

src/components/Modals/GettingStartedModal.tsx
jeff-phillips-18: I was trying to leave conflicts out of this PR. There are 2 `GettingStartedModal` components, one for app-studio, one for hacbs. When we do the work to resolve this we can combine the directories.

src/components/Components/__tests__/ComponentListView.spec.tsx
jeff-phillips-18: Trying hard not to mix in other types of rework/refactoring

karthikjeeyar:
LGTM! Seems like the e2e failures happens in all the open PRs

rohitkrai03 (changes requested):
@jeff-phillips-18 The `hacbs=false` flag doesn't work anymore for switching to old app studio UI. Apart from that everything seems to be working fine.

sahil143:
LGTM!

src/components/Modals/GettingStartedModal.tsx
```diff
@@ -44,7 +44,7 @@ type Level = {
   footerButton: { title: string; cta: any };
 };
 
-export const LOCAL_STORAGE_KEY = 'hacbs/getting-started-modal';
+export const LOCAL_STORAGE_KEY = 'getting-started-modal';
```
rottencandy, line 47: The app studio modal is also using the same local storage key.

karthikjeeyar: @rottencandy There is a ticket and a [PR](https://github.com/openshift/hac-dev/pull/285) to remove GettingStartedModals completely. So i think we should fine with this now

rottencandy: Okay sounds good

rohitkrai03: This might be the reason why e2e tests are getting stuck on getting started modal.

jeff-phillips-18: I can put it back for now and let it be removed if/when they are completely removed.

karthikjeeyar: > This might be the reason why e2e tests are getting stuck on getting started modal.

No, when i checked locally cypress is able to click on the Create Application button behind the modal and proceeded with the flow. I think the issue is, Applications is not loading in the ephemeral env, you can see this issue in Abhi's PR where he removed this modal completely but still e2e is failing with that error

image

PR #284, opened 2022-12-14
Use App studio Import flow

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/HAC-2539

## Description

Use Appstudio flow as the default flow. This flow allows the user to use git repo or a container image or samples and the component created will not contain `pipelinesascode` annotation.

## Type of change

- [x] Feature

ImportFlow

## How to test or reproduce?

1. Add Application and components (git source/container image or via samples) through Import flow wizard.

## Unit tests

```
 ImportForm
    ✓ should handle form reset (20 ms)
    ✓ should not set inAppContext if application name is not passed (4 ms)
    ✓ should set inAppContext if application name is passed (3 ms)
    ✓ should navigate to application page on form submit (5 ms)
    ✓ should warn the users about the errors on form submit (46 ms)
```

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge


karthikjeeyar:
cc: @Ranelim @MariaLeonova

christianvogt:
@karthikjeeyar we do want a wizard eventually to let the user opt into the flow where we set `pipelinesascode` on the component.

Why don't we add a temporary UI element that allows the user to create a component where we set `pipelinesascode`. This shouldn't happen for the samples flow.

Two suggestions:
- add another submit button
- add a checkbox

This is a temporary solution so which ever is easiest.

karthikjeeyar:
Updated e2e tests

Import flow test:

Screenshot 2022-12-15 at 7 39 11 PM

beaumorley:
Could we improve the copy for the PR checkbox? Here is a suggestion. 
[ ] Send pull request
This will create a custom pipeline in your repository.
FYI @megan-hall

karthikjeeyar:
Thanks @beaumorley I have updated the copy and screenshot.

integration-tests/utils/Applications.ts
```diff
@@ -39,10 +39,12 @@ export class Applications {
...
   }
 
-  static createComponent(publicGitRepo: string, componentName: string, integrationTestName: string, optionalForRelease: boolean = false) {
+  static createComponent(publicGitRepo: string, componentName: string) {
     addComponentStep(publicGitRepo);
     reviewComponentsStep(componentName);
-    createBuildStep();
+  }
+
+  static createIntegrationTest(integrationTestName: string, optionalForRelease: boolean = false) {
```
dheerajodha, line 47: Why did we create a new method `createIntegrationTest` here?

integration-tests/tests/add-component-to-application.spec.ts
```diff
@@ -16,15 +16,9 @@ describe('Create Components using the UI', () => {
     'https://github.com/nodeshift-starters/devfile-sample.git',
   ];
   const componentNames = ['java-quarkus', 'go', 'go-component', 'nodejs'];
```
dheerajodha, line 18: wondering if we still need to use these pre-defined component names when they now have randomly generated unique names instead. wdyt?

dheerajodha: That's a good point, I missed that.
> is that fine?

Yes sure

src/hacbs/components/ImportForm/ImportForm.tsx
```diff
@@ -44,30 +35,27 @@ const ImportForm: React.FunctionComponent<ImportFormProps> = ({ applicationName
...
-    onApplicationCreated: setApplicationData,
-    onComponentsCreated: () => setComponentsCreated(true),
-  });
+  const steps = useImportSteps(applicationName);
 
   const handleSubmit = React.useCallback(
-    ({ application, inAppContext }: FormValues) => {
-      const appName = inAppContext ? application : applicationDataRef.current?.metadata?.name;
-      navigate(`/stonesoup/applications/${appName}`);
+    (values: FormValues, formikHelpers: FormikHelpers<FormValues>) => {
```
rohitkrai03, line 41: Why keep this separate `ImportForm` component in hacbs folder? Why not just replace the original one with this?

rohitkrai03 (changes requested):

src/hacbs/components/ImportForm/ImportForm.tsx
karthikjeeyar: Moved contents from hacbs folder to src and removed unused files.

PR #305, opened 2023-01-02
Update hac ui tests

dheerajodha:
## Fixes 

https://issues.redhat.com/browse/HACBS-1558

## Description

Since the "Review Component" page has been removed from the import flow when we create a component using a "sample", there's no way to rename the uniquely generated component names which result in problems during assertions. Updated the tests to fetch the component names from the response of an API.

## Type of change

- [x] Bugfix

## How to test or reproduce?

Run the UI tests locally.

## Browser conformance: 

- [x] Chrome


karthikjeeyar (commented):
Thanks Dheeraj for fixing this, I had tested locally except private test flows all other are passing.

Screenshot 2023-01-03 at 2 11 56 PM

src/components/ImportForm/ReviewSection/ReviewComponentCard.tsx
```diff
@@ -65,6 +65,7 @@ export const ReviewComponentCard: React.FC<ReviewComponentCardProps> = ({
           'aria-label': name,
           'aria-labelledby': `review-${name} toggle-${name}`,
           'aria-expanded': expandedComponent,
+          'data-test': 'toggle-button',
```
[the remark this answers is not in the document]
dheerajodha: Done. To do this, I also had to update the `create-application-from-public-git-source.spec.ts` to use network calls for fetching component names. So, please check again, if possible.

karthikjeeyar:
> What about getting the name from the UI once it's shown to the user?

This is not be possible with samples flow as it does not have review section at the end to get the name from UI.

dheerajodha:
> This is not be possible with samples flow as it does not have review section at the end to get the name from UI.

Yes, this was the sole reason why we had to use the response body of APIs to fetch component names. And while I was fixing that test (`create-application-from-sample.spec.ts`), I thought maybe it's a good practice to fetch data from the API and use this data to assert what we see on the UI. In a way, that might cross-check the data sent by API.

But I'm totally open to what you folks think about this.

Katka92:
From my point of view, the UI tests should stick to the UI as much as possible. They should replicate what the user does as much as possible. So, if not needed, they don't touch API, cookies etc. I see the issue with the sample use case and I'm OK with this solution, but it seems redundant to me for the others.

dheerajodha:
> From my point of view, the UI tests should stick to the UI as much as possible. They should replicate what the user does as much as possible. So, if not needed, they don't touch API, cookies etc. I see the issue with the sample use case and I'm OK with this solution, but it seems redundant to me for the others.

I discussed this with Jan as well, and he agreed with what you're suggesting so I've removed the `cy.intercept()`s where they are not strictly needed. Plus I've fixed the 2 failing private tests that you noticed locally.

integration-tests/tests/create-application-from-public-git-source.spec.ts
```diff
@@ -54,9 +54,13 @@ describe('Create Component from Public Git Source', () => {
...
     it('Check Changing Resources', () => {
-      componentPage.expandDetails(componentName);
-      componentPage.setCpuByButton(cpuCount + 1, cpuUnit);
-      componentPage.setRam(ramValue, ramUnit);
+      cy.get(ComponentsPagePO.extractComponentName).then((innerText) => {
+        componentPage.componentName = innerText.text().trim();
+
+        componentPage.expandDetails(componentPage.componentName);
+        componentPage.setCpuByButton(cpuCount + 1, cpuUnit);
+        componentPage.setRam(ramValue, ramUnit);
```
Katka92, line 62: As the component name is now known outside the callback, I would move those lines outside. But that's more like my own feeling. If you want to leave it as it is, feel free.

integration-tests/utils/Applications.ts
```diff
@@ -45,12 +45,16 @@ export class Applications {
   }
 
 
-  static createdComponentExists(componentName: string, applicationName: string) {
+  static createdComponentExists(componentName: string, applicationName: string, strictChecking: boolean = false) {
```
Katka92, line 48: Is this strictChecking used somewhere?

integration-tests/tests/create-application-from-public-git-source.spec.ts
dheerajodha: I'd say we leave it here because I spent a good amount of time trying to move these lines outside and make it work but it fails. Since cypress commands are async the `componentName` variable would hold the `undefined` value outside of the `.then()` block.

integration-tests/utils/Applications.ts
dheerajodha: No, removed.

rohitkrai03:
@dheerajodha I tried running the e2e tests locally. All tests seem to be passing though I noticed in some tests the older application details page was being shown. Are we setting the `hacbs=false` feature flag somewhere in the test?

integration-tests/tests/create-application-from-sample.spec.ts
```diff
@@ -6,17 +6,27 @@ import { Applications } from '../utils/Applications';
...
+  ];
 
-  before(() => {
-    // Disable HACBS
-    localStorage.setItem('hacbs', 'false');
-    // Need to reload the page after enabling HACBS via localStorage
-    cy.reload();
+  beforeEach(function () {
+    localStorage.setItem(LOCAL_STORAGE_KEY_GS_MODAL, 'true');
+    localStorage.setItem(LOCAL_STORAGE_KEY_APPLICATION_MODAL, 'true');
```
Katka92, line 23: I see you set those in every test. What about setting it up once there https://github.com/dheerajodha/hac-dev/blob/fix-hacdev-tests/integration-tests/support/commands/hooks.ts#L15 ? It should work for all tests then.

integration-tests/tests/create-application-from-public-git-source.spec.ts
Katka92: Ok, thx for the explanation.

integration-tests/tests/create-application-from-sample.spec.ts
dheerajodha: Done, thank you!

dheerajodha:
@rohitkrai03 Ready to merge? The PR check env is still not fixed so its unreliable.

rohitkrai03 (commented):
/approve

PR #310, opened 2023-01-09
Follow CR name requirements for application input field

rottencandy:
## Fixes 
https://issues.redhat.com/browse/HAC-2353
https://issues.redhat.com/browse/HAC-3048

## Description
- Use CR name for field instead of displayName
- Change field label to "Name"
- Change placeholder to "my-first-app" and increment suffix if name is already used
- Show Inline errors for incorrect name format
- Use RFC compliant regex for resource name fields

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

https://user-images.githubusercontent.com/20013884/216340465-687b8791-050d-4a68-889a-e281a33c618c.mp4

## How to test or reproduce?
Create applications from import form.

## Browser conformance: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/ImportForm/ApplicationSection/ApplicationSection.tsx
+++ b/src/components/ImportForm/ApplicationSection/ApplicationSection.tsx
@@ -1,9 +1,26 @@
 import * as React from 'react';
 import { TextContent, Text, HelperText, FormSection } from '@patternfly/react-core';
+import { useField } from 'formik';
 import { InputField } from '../../../shared';
 import { HeadTitle } from '../../HeadTitle';
+import { useValidApplicationName } from '../utils/useValidApplicationName';
 
 const ApplicationSection: React.FunctionComponent = () => {
+  const [validName] = useValidApplicationName();
+  const [, { value, touched }, { setValue, setTouched }] = useField<string>('application');
+
+  React.useEffect(() => {
+    if (!value && !touched && validName) {
+      setValue(validName);
+    }
+  }, [setValue, touched, validName, value]);
+
+  React.useEffect(() => {
+    if (value && !touched) {
+      setTouched(true, false);
+    }
+  }, [setTouched, touched, value]);
+
   return (
     <>
       <HeadTitle>Import - Name application | CI/CD</HeadTitle>
@@ -12,12 +29,7 @@ const ApplicationSection: React.FunctionComponent = () => {
         <HelperText>Enter a name for your application.</HelperText>
       </TextContent>
       <FormSection>
-        <InputField
-          name="application"
-          label="Application name"
-          placeholder="My Application"
-          required
-        />
+        <InputField name="application" label="Name" placeholder="Enter name" required />
       </FormSection>
     </>
   );
--- a/src/components/ImportForm/utils/useValidApplicationName.ts
+++ b/src/components/ImportForm/utils/useValidApplicationName.ts
@@ -1,0 +1,32 @@
+import * as React from 'react';
+import { useK8sWatchResource } from '@openshift/dynamic-plugin-sdk-utils';
+import { ApplicationGroupVersionKind } from '../../../models';
+import { ApplicationKind } from '../../../types';
+import { useNamespace } from '../../../utils/namespace-context-utils';
+
+const BASE_NAME = 'my-app';
+
+export const useValidApplicationName = (): [string, boolean, unknown] => {
+  const namespace = useNamespace();
+
+  const [applications, loaded, loadErr] = useK8sWatchResource<ApplicationKind[]>({
+    groupVersionKind: ApplicationGroupVersionKind,
+    namespace,
+    isList: true,
+  });
+
+  const validName = React.useMemo(() => {
+    if (!loaded) {
+      return;
+    }
+    let name = BASE_NAME,
+      i = 1;
+    while (applications.some((app) => app.metadata.name === name)) {
+      name = `${BASE_NAME}-${i}`;
+      i++;
+    }
+    return name;
+  }, [applications, loaded]);
+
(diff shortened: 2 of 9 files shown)


src/components/ImportForm/ApplicationSection/ApplicationSection.tsx
```diff
@@ -1,21 +1,19 @@
 import * as React from 'react';
 import { TextContent, Text, HelperText, FormSection } from '@patternfly/react-core';
 import { InputField } from '../../../shared';
+import { useValidApplicationName } from '../utils/useValidApplicationName';
 
 const ApplicationSection: React.FunctionComponent = () => {
+  const [validName] = useValidApplicationName();
```
rohitkrai03, line 7: Do we actually need to do a validation for just placeholder text? We are already doing that on next.

rohitkrai03 (changes requested):

src/components/ImportForm/ApplicationSection/ApplicationSection.tsx
rottencandy: The hook returns an incremented name that isn't used, like `my-first-app-$n`. UX wanted the placeholder value to increment https://issues.redhat.com/browse/HAC-2353

rohitkrai03:
Hey, I've merged the PR that fixed the e2e tests - https://github.com/openshift/hac-dev/pull/305. So going forward it would be good to run the e2e tests locally and paste the screenshot here on the PR before merging. Can you do that?

src/components/ImportForm/ApplicationSection/ApplicationSection.tsx
rohitkrai03: How does it adds value if this is just a placeholder text? It just unnecessarily slows the process for the user. It would have made sense when we were actually recommending the application in the input field and not just showing a placeholder text which eventually needs to be entered by the user. @beaumorley What do you think?

rohitkrai03: @beaumorley ^^

Misjohns: @rohitkrai03 @beaumorley The placeholder text seems weird to me. Can we automatically populate the field with "my-first-app" (reg text not placeholder) and the placeholder "Enter name" only displays if the user removes all the text? So if user comes here they can simply hit enter.

rottencandy: @Misjohns updated!

https://user-images.githubusercontent.com/20013884/214018741-57b9b201-0caf-4f6c-ab81-6fdf73d67505.mp4

rohitkrai03: @rottencandy Do we need the `onNext` validation of application name anymore? i think we can remove that check if the application name is always recommended based on what can be created.

rohitkrai03 (changes requested):

src/components/ImportForm/ApplicationSection/ApplicationSection.tsx
rottencandy: But we still allow users to change to a different name, should we not do a check on the field in that case?

rohitkrai03: Well, yeah in that case we still need that check.

src/components/ImportForm/utils/useValidApplicationName.ts
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import { useK8sWatchResource } from '@openshift/dynamic-plugin-sdk-utils';
+import { ApplicationGroupVersionKind } from '../../../models';
+import { ApplicationKind } from '../../../types';
+import { useNamespace } from '../../../utils/namespace-context-utils';
+
+const BASE_NAME = 'my-first-app';
```
rohitkrai03, line 7: This will always result in `my-first-app-${some number}` even if I have multiple apps already created. Would be better to use generic name like `my-app` or something. Or if we want to be better at recommending name then we can say `my-first-app`, `my-second-app`, etc based on the number of applications already present.

rohitkrai03 (changes requested):

rohitkrai03 (changes requested):
Seeing something weird when you go to next step, come back and try to delete the name. 

https://user-images.githubusercontent.com/6041994/216282551-1446f3fc-dd45-4f2e-8a6d-6e2e41d6d539.mov

src/components/ImportForm/utils/useValidApplicationName.ts
rohitkrai03: I think we can go ahead with simple implementation of `my-app-1`, `my-app-2`, etc for MVP.

rottencandy: Makes sense, updated.

rottencandy:
> Seeing something weird when you go to next step, come back and try to delete the name.

Old formik value + touching was causing this issue, fixed.

https://user-images.githubusercontent.com/20013884/216340465-687b8791-050d-4a68-889a-e281a33c618c.mp4

rohitkrai03 (approved):
/approve

rottencandy:
Removing value check is causing CI to fail, looking into it

beaumorley:
We discussed this today on our sync as a good MVP solution. LGTM.  

@Misjohns is going to be out tomorrow. So may not respond.

rohitkrai03 (approved):
/lgtm

PR #329, opened 2023-01-16
Fix tests.

Katka92:
## Fixes 
Differents fixes to be able to have green PR check again.

## Description
As the environment was not ready for quite some time, we must clean and polish our tests to run well there now. This is what this PR is serving for.


rohitkrai03 (approved):
/approve

PR #336, opened 2023-01-19
fix environment variables deletion while updating a component

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/HAC-2959

## Description

Deleting environment variables from UI was not reflecting in the components CR.

## Type of change

- [x] Bugfix

https://user-images.githubusercontent.com/9964343/213451990-367c5587-13fb-4e97-bbe4-5fbcf01e6457.mov

## Unit tests

```
Create Utils
    ✓ Should delete the environment variables while updating (1 ms)
    ✓ Should update the environment variables with new values (1 ms)
```

## How to test or reproduce?

1. While creating components, Add environment variables under advanced deployment options 
2. Go to component setting page from component list view and remove all the environment variables under advanced options.

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

cc: @christianvogt

### DIFF
--- a/src/utils/create-utils.ts
+++ b/src/utils/create-utils.ts
@@ -5,7 +5,6 @@ import {
 } from '@openshift/dynamic-plugin-sdk-utils';
 import isEqual from 'lodash/isEqual';
 import isNumber from 'lodash/isNumber';
-import merge from 'lodash/merge';
 import uniqueId from 'lodash/uniqueId';
 import {
   ApplicationModel,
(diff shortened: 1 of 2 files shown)


src/utils/create-utils.ts
```diff
@@ -114,7 +114,12 @@ export const createComponent = (
   };
 
   const resource =
-    verb === 'update' ? merge({}, originalComponent || {}, newComponent) : newComponent;
+    verb === 'update'
+      ? {
+          ...merge({}, originalComponent || {}, newComponent),
+          spec: newComponent.spec,
```
sahil143, line 120: @karthikjeeyar This will completely replace the spec of the original component instead of a partial update.

christianvogt: I think that's fair.
Why not just do:
```
  const resource =
    verb === 'update' ? { ...originalComponent, spec: newComponent.spec } : newComponent;
```

karthikjeeyar: Updated the code.

rohitkrai03 (approved):
/lgtm

PR #365, opened 2023-01-26
add plr results to details page

christianvogt:
## Fixes 
TBD

## Description
Displays a table of the key value pair of results for the pipeline run on the details page.
If a URL is present in the data, it is converted to a link.
For the most part, the code was taken from the developer console.

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

When results are present, a new table is shown at the bottom of the page:
image

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/utils/render-utils.tsx
+++ b/src/utils/render-utils.tsx
@@ -1,0 +1,25 @@
+import * as React from 'react';
+import ExternalLink from '../shared/components/links/ExternalLink';
+
+const URL_REGEXP =
+  /https?:\/\/(?:www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b(?:[-a-zA-Z0-9()@:%_+.~#?&/=]*)/;
+export const GROUP_MATCH_REGEXP = new RegExp(`^(.*\\s)?(${URL_REGEXP.source})(\\s.*)?$`, 'i');
+
+export const handleURLs = (value: string): React.ReactNode => {
+  if (typeof value !== 'string') return value;
+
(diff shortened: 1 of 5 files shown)


src/utils/render-utils.tsx
```diff
@@ -0,0 +1,25 @@
...
+import ExternalLink from '../shared/components/links/ExternalLink';
+
+const URL_REGEXP =
+  /https?:\/\/(?:www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b(?:[-a-zA-Z0-9()@:%_+.~#?&/=]*)/;
+export const GROUP_MATCH_REGEXP = new RegExp(`^(.*\\s)?(${URL_REGEXP.source})(\\s.*)?$`, 'i');
+
+export const handleURLs = (value: string): React.ReactNode => {
+  if (typeof value !== 'string') return value;
+
+  const matches = value.match(GROUP_MATCH_REGEXP);
```
karthikjeeyar, line 11: This regex covers only `https` urls, shall we include regex to match `quay.io` as well, so that we can link it to the quay registry?

image

christianvogt: @karthikjeeyar I think this may be worth looking into via a separate issue because changing this into a URL doesn't actually bring us to a useful page.

karthikjeeyar: Ah I see, tag page does not exist.

christianvogt:
@karthikjeeyar rebased

PR #371, opened 2023-01-30
Added whats next section to application overview

rohitkrai03:
## Fixes 

- https://issues.redhat.com/browse/HAC-2686

## Description

- Added what's next section to application overview page.
- Using random help topic id for all learn more links. Will be updating it with correct ones when docs create them.
- Need to add some unit tests.

## Type of change

- [x] Feature

## Screen shots / Gifs for design review 

https://user-images.githubusercontent.com/6041994/215546704-16e99091-d373-4bc9-b613-898ef288b9d6.mov

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

PR #372, opened 2023-01-30
update all empty state copy / icons

jeff-phillips-18:
## Fixes 
Fixes [HAC-2837](https://issues.redhat.com/browse/HAC-2837)

## Description
Updates all copy and icons for all empty states per https://docs.google.com/document/d/1ouUwFdxl0f4j0sO5Zi1JwQZvnk2JhzNhwVkOo_JqxbY

Removed the unused old version of the Application details view in order to more easily adjust the empty state for the view.

## Type of change
- [x] Feature

image

image

image

image

image

image


Misjohns:
@jeff-phillips-18 Maybe you are still working to update the microcopy but the images above don't include the updated copy. Also, we need to add some space to the top and bottom of the message. Let's use 48px padding from the top of the dotted container and the top of the icon and 48px padding from the bottom of the buttons and the dotted container.

jeff-phillips-18:
> Let's use 48px padding from the top of the dotted container and the top of the icon and 48px padding from the bottom of the buttons and the dotted container.

@Misjohns We can, but currently this is using standard PatternFly spacing.

jeff-phillips-18:
@Misjohns Updated and updated the images above.

rohitkrai03 (changes requested):
I am using SVGs provided by UX in this https://github.com/openshift/hac-dev/pull/371 instead of pngs. I think it would be better to use SVGs. What do you think?

jeff-phillips-18:
> I am using SVGs provided by UX in this #371 instead of pngs. I think it would be better to use SVGs. What do you think?

SVGs are preferable. Do we have SVGs for these icons?

jeff-phillips-18:
Updated to use SVGs

jeff-phillips-18:
@Misjohns Can you please take another look and comment?

christianvogt:
Looks like the pipeline runs filter list wasn't update with the new icon:

image

christianvogt:
Other than the above, it seems everything else is in order.

/approve

christianvogt:
Also the samples page could use a new icon:
image

jeff-phillips-18:
Updated the remain empty states

PR #376, opened 2023-01-31
fix last deploy time in application list view

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/HAC-2207

## Description

The `Last deploy` column of the application list view renders the application's `creationTimestamp` as the value. This is misleading as it does not represent the last deploy timestamp.

**Solution:**

Using snapshotEnvironmentBinding cr's `.status.componentDeploymentConditions[0].lastTransitionTime` value to fetch the latest deployed time, however this is not a guaranteed/accurate at the moment so we have a backend ticket to improve the accuracy of the lastDeployed time https://issues.redhat.com/browse/GITOPSRVCE-375 till then we can use the `lastTransitionTime` to get the timestamp.

## Type of change

- [x] Bugfix

Screenshot 2023-01-31 at 6 48 33 PM image

---
Screenshot 2023-01-31 at 6 48 42 PM

## How to test or reproduce?

Check the last deployed timestamp in the application list view and environment list view.

## Unit tests

```
 Application List Row
    ✓ should render the last deployed environment timestamp in application list (11 ms)
```

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

cc: @jgwest @christianvogt

### DIFF
--- a/src/components/ApplicationListView/ApplicationListRow.tsx
+++ b/src/components/ApplicationListView/ApplicationListRow.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { Link } from 'react-router-dom';
 import { useK8sWatchResource } from '@openshift/dynamic-plugin-sdk-utils';
 import { pluralize, Skeleton } from '@patternfly/react-core';
+import { useAllApplicationEnvironmentsWithHealthStatus } from '../../hooks/useAllApplicationEnvironmentsWithHealthStatus';
 import { ComponentGroupVersionKind } from '../../models';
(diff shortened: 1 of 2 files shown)


christianvogt:
When we have multiple environments per application, the last deploy on the application list page is sort of arbitrary. We should consult with UX on how to better convey what this means as we probably want to say what env was the last deploy on or give some sort of context. If there are 5 envs for an application. The last deploy isn't that useful.

PR #380, opened 2023-02-01
Use sample component names as retrieved from CDQ

jeff-phillips-18:
## Fixes 
Fixes [HAC-3024](https://issues.redhat.com/browse/HAC-3024)

## Description
When creating sample components, use the suggested name returned by CDQ rather than prepending the application name and adding the `-sample` suffix.

## Type of change
- [x] Bugfix


rohitkrai03 (approved):
/lgtm

PR #382, opened 2023-02-01
Fix to hide action menus when bringing up another

jeff-phillips-18:
## Fixes 
Fixes [HAC-3072](https://issues.redhat.com/browse/HAC-3072)

## Description
Action menu click was not propagating the event so other action menus did not close. Change to propagate the event, but delay the menu open so that it doesn't automatically close when catching the click event.

## Type of change
- [x] Bugfix


rohitkrai03 (approved):
/lgtm

PR #394, opened 2023-02-03
Add workspace switcher

rottencandy:
## Fixes 
https://issues.redhat.com/browse/HAC-2747

## Description
- Add workspace switcher using the ContextSwitcher component.
- Update ContextSwitcher component to support recursive menu items
- Update breadcrumb to support custom components

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

https://user-images.githubusercontent.com/20013884/216656542-c1b39f26-629a-4338-b42f-45e724a11c9a.mp4

## Browser conformance: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/ApplicationDetails/ApplicationDetails.tsx
+++ b/src/components/ApplicationDetails/ApplicationDetails.tsx
@@ -2,7 +2,8 @@ import React from 'react';
 import { Link, useNavigate } from 'react-router-dom';
 import { useFeatureFlag } from '@openshift/dynamic-plugin-sdk';
 import { useK8sWatchResource } from '@openshift/dynamic-plugin-sdk-utils';
-import { Bullseye, Spinner } from '@patternfly/react-core';
+import { Badge, BreadcrumbItem, Bullseye, Spinner } from '@patternfly/react-core';
+import styles from '@patternfly/react-styles/css/components/Breadcrumb/breadcrumb';
 import { useChrome } from '@redhat-cloud-services/frontend-components/useChrome';
 import { ApplicationGroupVersionKind } from '../../models';
 import { HttpError } from '../../shared/utils/error/http-error';
@@ -23,6 +24,7 @@ import ApplicationOverviewTab from './tabs/ApplicationOverviewTab';
 import ComponentsTab from './tabs/ComponentsTab';
 import EnvironmentsTab from './tabs/EnvironmentsTab';
 import IntegrationTestsTab from './tabs/IntegrationTestsTab';
+import { WorkspaceSwitcher } from './WorkspaceSwitcher';
 
 import './ApplicationDetails.scss';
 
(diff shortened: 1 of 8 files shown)


rottencandy:
cc @openshift/team-hacdev-ux @karthikjeeyar

itsptk:
@rottencandy Thanks for tagging. I think we'd just want to remove the arrow ">" between the (WS) label and the workspace name, and then also replace the arrow ">" between the workspace switcher icon button and the 'proper' breadcrumb with some other separator. The design has a "/" line but we can work through whatever works.

image

rottencandy:
@itsptk does this look good? I couldn't find that icon in PF libs so I used a slash character instead

itsptk:
> @itsptk does this look good? I couldn't find that icon in PF libs so I used a slash character instead

Thanks @rottencandy , this looks better. This might be a change to the ContextSwitcher but would it be possible to reduce the amount of space between the workspace/application name and the icon button to open the switcher? Right now there is more space between the resource and the icon button than there is the separator after it, so they sort of look disassociated. 

I think that slash looks good for now, we may take a look at another separator option and get back to you this week. Thanks again.

karthikjeeyar:
In the breadcrumbs `Application` contains a broken link.

Screenshot 2023-02-08 at 12 10 18 PM

rottencandy:
Updated!

itsptk:
> Updated! 

This looks really good, thanks @rottencandy . If possible, could we increase the space between the separator and the start of the breadcrumb (the space between "/" and "Applications" here) so that the separation between the workspace switcher and  breadcrumb is more apparent? Maybe closer matching the distance between the "..." icon button and the "/" seperator?

Also, we talked through it with the UX team more and are thinking we'd like to switch the separator from "/" to "|" to maybe make it slightly clearer. Thanks very much!

rottencandy:
Switched to `│` and added equal spacing to both sides, looks like this now:


cc @karthikjeeyar

rohitkrai03 (approved):
/lgtm

src/components/ApplicationDetails/ApplicationDetails.tsx
```diff
@@ -82,14 +87,31 @@ const ApplicationDetails: React.FC<HacbsApplicationDetailsProps> = ({ applicatio
         data-test="application-details-test-id"
         headTitle={appDisplayName}
         breadcrumbs={[
-          { path: '/stonesoup/applications', name: 'Applications' },
+          <Badge key="badge" isRead>
```
rohitkrai03, line 90: @rottencandy Any reason to switch from breadcrumb items to breadcrumbs only?

rottencandy: The `breadcrumbitems` prop would only allow adding components to the end of breadcrumbs list, and the `breadcrumbs` prop did not support custom item components.

I added support for custom components to `breadcrumbs` so we could use it to render both paths and custom items at any position in the breadcrumbs list. And we wouldn't need `breadcrumbiems` anymore.

rohitkrai03 (approved):
/lgtm

rohitkrai03:
/hold cancel

We've put this feature behind MVP feature flag. It will be removed in the follow up https://github.com/openshift/hac-dev/pull/406

PR #401, opened 2023-02-07
Fix to show error state rather than spinner

jeff-phillips-18:
## Fixes 
Fixes [HAC-3050](https://issues.redhat.com/browse/HAC-3050)

## Description
Handle errors at top level pages and display an error state if an error is returned while loading the data.
Created an ErrorState component that handles showing errors consistently, and generically handles 404 errors.

## Type of change
- [x] Bugfix

image

/cc @MariaLeonova

### DIFF
--- a/config/remotePlugin.js
+++ b/config/remotePlugin.js
@@ -324,7 +324,7 @@ const routeExtensions = [
     },
   },
   {
-    type: 'console.page/route',
+    type: 'core.page/route',
     properties: {
       path: '/stonesoup/workspace-settings/environment/create',
       exact: true,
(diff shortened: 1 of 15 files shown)


MariaLeonova:
Thank you, @jeff-phillips-18 ! This looks great!

config/remotePlugin.js
```diff
@@ -349,6 +349,32 @@ const routeExtensions = [
...
+      path: '/stonesoup',
+      exact: false,
+      component: {
+        $codeRef: 'NotFound',
+      },
+    },
+    flags: {
+      required: ['SIGNUP'],
+    },
+  },
```
rohitkrai03, line 364: We don't need to add the deprecated `console` extensions anymore.

rohitkrai03 (changes requested):

config/remotePlugin.js
jeff-phillips-18: Shall I remove the others?

christianvogt: @jeff-phillips-18 may as well if they are redundant and unnecessary.

PR #405, opened 2023-02-08
Update build node description in lifecycle visualization

rohitkrai03:
## Fixes 

- https://issues.redhat.com/browse/HAC-2914

## Description

- Updated build node description in lifecycle visualization.

## Type of change

- [x] Bugfix

Screenshot 2023-02-08 at 10 26 41 PM

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/ApplicationDetails/tabs/overview/visualization/utils/node-utils.ts
+++ b/src/components/ApplicationDetails/tabs/overview/visualization/utils/node-utils.ts
@@ -31,7 +31,7 @@ const RUN_STATUS_SEVERITIES = [
 
 const COMPONENT_DESC =
   'A component is an image built from code in a source repository. Applications are sets of components that run together on environments.';
-const BUILD_DESC = `Every component requires a build in order to deploy it to an environment. Whenever you create a new component, we send a pull request to merge our build pipeline to your component's source code`;
+const BUILD_DESC = `Every component requires a build to deploy to an environment. If you choose to use a default pipeline, we'll automatically trigger a single build. Subsequent builds will need to be manually requested. If you choose to customize your build pipeline, you'll walk through a few steps to opt in, including merging a pull request with the default pipeline definition. We'll generate new builds for each merged pull request.`;
 const TESTS_DESC = `An integration test is a test pipeline you build and add to the application CI/CD definition. It tests each component as it’s added and then tests the entire application as a whole.`;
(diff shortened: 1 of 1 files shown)


src/components/ApplicationDetails/tabs/overview/visualization/utils/node-utils.ts
```diff
@@ -31,7 +31,7 @@ const RUN_STATUS_SEVERITIES = [
 
 const COMPONENT_DESC =
   'A component is an image built from code in a source repository. Applications are sets of components that run together on environments.';
-const BUILD_DESC = `Every component requires a build in order to deploy it to an environment. Whenever you create a new component, we send a pull request to merge our build pipeline to your component's source code`;
+const BUILD_DESC = `Every component requires a build to deploy it to an environment. If you're using a default pipeline, the build will be triggered automatically. If you choose to customize your build pipeline, we'll send you a pull request.`;
```
abigaeljamie, line 34:
```suggestion
const BUILD_DESC = `Every component requires a build to deploy to an environment. If you're using a default pipeline, the build will be triggered automatically. If you choose to customize your build pipeline, we'll send you a pull request.`;
```

abigaeljamie (commented):
Is it OK if we remove the "it"? After reading it over, someone on the team requested that we remove "it," and it sounds a bit smoother :)

abigaeljamie:
@christianvogt , did you want to give this description a quick review?

christianvogt:
@abigaeljamie the message is confusing for the customized pipelines. In order to opt into customize a pipeline the user must perform several states, one of which is receiving a pull request with the default pipeline definition and then merging it.

However once that is complete, new builds are generated per PR and merge. This messaging makes it sound like we will send a pull request for every build.

Furthermore, for the default scenario, the user only gets a single build triggered automatically. Subsequent builds must be manually requested.

abigaeljamie:
Thank you for that context, @christianvogt ! We can lengthen the message a bit to include more information about that technical functionality. What do you think of this? 

_Every component requires a build to deploy to an environment. If you choose to use a default pipeline, we'll automatically trigger a single build. Subsequent builds will need to be manually requested. If you choose to customize your build pipeline, you'll walk through a few steps to opt in, including merging a pull request with the default pipeline definition. We'll generate new builds for each merged pull request._

abigaeljamie (approved):
Thank you! LGTM

PR #408, opened 2023-02-08
Update SDK to V2 in order to support setting/fetching workspaces

florkbr:
@karthikjeeyar @rohitkrai03 please review. This should unblock us to use the new `useWorkspace` hook and setting/fetching workspace from localstorage. 

Tested locally and _seems_ safe and functioning: 
image


package.json
```diff
@@ -66,7 +66,6 @@
     "@openshift/dynamic-plugin-sdk-webpack": "1.0.0-alpha3",
```
florkbr, line 66: FWIW this is the package that should be updated ASAP (see todo here: https://github.com/openshift/hac-dev/pull/408/files#diff-4be71c3182a0de045ef56aec0e1aa8d5bf9075135637452164bda6e0cb015aa0R17)

I attempted to do so but ran into an issue where version `^2.0.0` of `@openshift/dynamic-plugin-sdk-webpack` requires `webpack` `^5.75.0` as a peer dependency,  however `webpack` comes from `@redhat-cloud-services/frontend-components-config` in hac-dev. This means  `@redhat-cloud-services/frontend-components-config` must be updated to the latest version. However in doing so, the production webpack build never completes and hangs indefinitely traversing the `node_modules` dir. I suspect it may be an issue with the sass loader.

florkbr:
Should resolve https://issues.redhat.com/browse/RHCLOUD-24171, but will need a separate issue tracking just the webpack portion of the upgrade.

package.json
```diff
@@ -26,9 +26,9 @@
   },
   "dependencies": {
     "@fortawesome/fontawesome-free": "^5.9.0",
-    "@openshift/dynamic-plugin-sdk": "^1.0.0",
-    "@openshift/dynamic-plugin-sdk-extensions": "^1.0.0",
-    "@openshift/dynamic-plugin-sdk-utils": "^1.0.0",
+    "@openshift/dynamic-plugin-sdk": "^2.0.1",
+    "@openshift/dynamic-plugin-sdk-extensions": "^1.1.0",
+    "@openshift/dynamic-plugin-sdk-utils": "^1.1.1",
```
karthikjeeyar, line 31: It errors out if I import and use the `useWorkspace` hook. After some investigation, it looks like
Webpack always servers 1.1.0 version of the utils (used in hac-core's [package.json ](https://github.com/openshift/hac-core/blob/main/frontend/package.json#L31)) and `dynamic-plugin-sdk-utils` is declared as a shared singleton module [hac-core/plugin.ts](https://github.com/openshift/hac-core/blob/main/frontend/config/plugins.js#L20), so we might need to update in hac-core as well?

loads old/1.1.0 version:

image

karthikjeeyar:
To fix the `ci/prow/test` failure, you need to mock the `commonFetch` function.

```
jest.mock('@openshift/dynamic-plugin-sdk-utils', () => ({
  commonFetch: jest.fn(),
}));
```
You could add this in `src/hacbs/components/Environment/__tests__/EnvironmentCard.spec.tsx`

package.json
rottencandy: For me it crashes even without importing the hook
```
ErrorWithCause: Failed to load module 'Applications' in extension hac-dev[19]_e12
...
Caused by: Error: Shared module redux doesn't exist in shared scope default
```
screenshot-1675941997

karthikjeeyar: @rottencandy redux issue is fixed already - [slack thread](https://redhat-internal.slack.com/archives/C038DJAP7HR/p1675938562913009?thread_ts=1675935775.289279&cid=C038DJAP7HR).

florkbr: PR here: https://github.com/openshift/hac-core/pull/132

src/hacbs/components/Environment/__tests__/EnvironmentCard.spec.tsx
```diff
@@ -10,6 +10,10 @@ jest.mock('@openshift/dynamic-plugin-sdk', () => ({
   useFeatureFlag: jest.fn(),
 }));
 
+jest.mock('@openshift/dynamic-plugin-sdk-utils', () => ({
+  commonFetch: jest.fn(),
```
florkbr, line 14: I'm not sure why this test suddenly started failing with my PR here. The SDK update should not impact `commonFetch` - it is still exposed through lib-utils: https://github.com/openshift/dynamic-plugin-sdk/blob/main/packages/lib-utils/src/index.ts#L15

Was this broken in main?

karthikjeeyar: No, it wasn't breaking in main.

florkbr:
Not sure if this error is related at all to the build/test failures: `15:21:34 /container_workspace/server_config_gen.sh: line 16: TRAVIS_BRANCH: unbound variable`...

florkbr:
I suspect failure is related to https://gitlab.cee.redhat.com/insights-platform/frontend-build-container/-/merge_requests/23. It seems server_config_gen.sh now defaults to caddy and not nginx

florkbr:
PR check failures are unrelated to this PR. See explanation here: https://github.com/openshift/hac-dev/pull/416

PR #419, opened 2023-02-10
fix crash when import a private Git Repository

christianvogt:
## Fixes 

https://issues.redhat.com/browse/HAC-3029

## Description

This issue only presents itself in production because the error is due to CORS.

If the network request to SPI fails, the UI was setting the error object into the state. We then try to render the error object which causes react to blow up. The solution is to extract the error message and set a string to the error state.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

No change.

## How to test or reproduce?
- go to create a component
- enter any git url that's unreachable
- manually enter a token
- press connect

You will need to force an error to occur in the network request to SPI in order to reproduce this.

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


rohitkrai03 (approved):
/lgtm

PR #421, opened 2023-02-13
Fix e2e test instability

jrichter1:
fix UI tests getting 404 because the application name sometimes gets mixed up with pre-generated value


rohitkrai03 (approved):
/lgtm

PR #428, opened 2023-02-17
Update GitHub links

rohitkrai03:
## Fixes 

- https://issues.redhat.com/browse/HAC-3174

## Description

- The Github App link needs to be dynamic based on what environment the UI is running on.
  - For ex - staging cluster will be using the new staging github app and prod cluster will use prod github app.
- The PR links also need to be updated based on the github app names.
- This PR updates the code to use `getEnvironment` from chrome API and use it to show correct github app links.

## Type of change

- [x] Feature

NO UI change.

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


src/hooks/useStoneSoupGitHubApp.ts
```diff
@@ -0,0 +1,38 @@
...
+  url: string;
+  name: string;
+};
+
+export const StoneSoupGitHubAppData: { [env: string]: StoneSoupGitHubAppDataType } = {
+  dev: { url: 'https://github.com/apps/appstudio-staging-ci', name: 'appstudio-staging-ci' },
+  stage: {
+    url: 'https://github.com/apps/red-hat-appstudio-ci-cd-staging',
+    name: 'red-hat-appstudio-ci-cd-staging',
+  },
```
karthikjeeyar, line 20: If I am not wrong  chrome will treat `console.dev.redhat.com` as `dev` environment, in that case we should interchange this dev and stage values. 


```suggestion
  stage: { url: 'https://github.com/apps/appstudio-staging-ci', name: 'appstudio-staging-ci' },
  dev: {
    url: 'https://github.com/apps/red-hat-appstudio-ci-cd-staging',
    name: 'red-hat-appstudio-ci-cd-staging',
  },
```

rohitkrai03: I added a dev value because eventually we might have a dev github app if there's a third cluster called dev. And I am using dev (old staging github app) as a default case in the hook so that it works for local dev as well as any other environment value that we do not know.

rohitkrai03: Stage is actually named `stage` because of the new staging cluster that the github app talks to.

PR #434, opened 2023-02-20
Filter out items flagged for deletion

jeff-phillips-18:
## Fixes 
Fixes [HAC-2642](https://issues.redhat.com/browse/HAC-2642)
Fixes [HAC-2348](https://issues.redhat.com/browse/HAC-2348)
Fixes [HAC-2912](https://issues.redhat.com/browse/HAC-2912)

## Description
Filter out items that are marked for deletion (deletionTimestamp is set). In the future, we should show the items but indicate that they are in the process of being deleted. See [HAC-2912](https://issues.redhat.com/browse/HAC-2912)

Also removed duplicate pipeline types from src/shared

## Type of change
- [x] Bugfix


karthikjeeyar (commented):
It looks the new libraries were added in integration-tests/package.json a few weeks ago in another PR, but package-lock.json was not pushed/updated to main branch.

/approve

src/hooks/usePipelineRunsForApplication.ts
```diff
@@ -3,10 +3,24 @@ import { useK8sWatchResource } from '@openshift/dynamic-plugin-sdk-utils';
...
-import { ComponentKind } from '../types';
+import { ComponentKind, PipelineRunKind } from '../types';
 import { useWorkspaceInfo } from '../utils/workspace-context-utils';
 
+export const usePipelineRun = (pipelineRunName: string): [PipelineRunKind, boolean, unknown] => {
+  const { namespace } = useWorkspaceInfo();
+  const [pipelineRun, loaded, error] = useK8sWatchResource<PipelineRunKind>({
+    groupVersionKind: PipelineRunGroupVersionKind,
+    name: pipelineRunName,
+    namespace,
```
christianvogt, line 14: I prefer if this hook relied on the namespace being passed in so it's more reusable in cases where later we have access to the namespace already.

It also makes the api more consistent with the other hooks.

christianvogt (commented):
overall it looks good but I haven't run the changes yet

PR #436, opened 2023-02-21
Fix environment UI crashing issue

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/HAC-3202

## Description

Environment page is breaking when the component creation fails with Gitops error then the SnapshotEnvironmentBinding's status is missing `ComponentDeploymentConditions` field. 

Note:
This is not happening in staging, development clusters. This issue was noticed when deploying frontend + backend components in quicklab clusters.

## Type of change

- [x] Bugfix

## How to test or reproduce?

This issue was noticed when deploying frontend + backend components in quicklab clusters.

https://github.com/ch007m/my-stonesoup/blob/main/howto-play.md#install-the-frontend

## Unit tests

```
  useAllApplicationEnvironmentsWithHealthStatus
    ✓ should return empty array if some resources are not loaded (12 ms)
    ✓ should return default status for managed environment (3 ms)
    ✓ should return Missing status (2 ms)
    ✓ should return Progressing status (2 ms)
    ✓ should return Healthy status (2 ms)
    ✓ should return Degraded status (2 ms)
    ✓ should return default values for partial status (2 ms)
```    

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

cc: @rohitkrai03

### DIFF
--- a/src/components/ApplicationListView/ApplicationListRow.tsx
+++ b/src/components/ApplicationListView/ApplicationListRow.tsx
@@ -22,7 +22,7 @@ const ApplicationListRow: React.FC<RowFunctionArgs<ApplicationKind>> = ({ obj })
       environmentsLoaded
         ? allEnvironments?.sort?.(
             (a, b) => new Date(b.lastDeploy).getTime() - new Date(a.lastDeploy).getTime(),
-          )?.[0].lastDeploy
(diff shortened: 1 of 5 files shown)


rohitkrai03 (approved):
/lgtm

PR #440, opened 2023-02-22
Force release the namespace.

Katka92:
## Fixes 
The bonfire namespace is not released after the tests.

## Description
The bonfire release command without the `-f` flag was waiting for the confirmation and so the namespace was not released after the test run but after the timeout.
Should be done as part of [HAC-3491](https://issues.redhat.com/browse/HAC-3491)


rohitkrai03 (approved):
/approve

PR #444, opened 2023-02-23
Show task run test output status on task run nodes

jeff-phillips-18:
## Fixes 
Fixes [HAC-3200](https://issues.redhat.com/browse/HAC-3200)

## Description
Determines the test status for pipeline run tasks and if the task run is successful, show any test status to indicate the test status issues.

## Type of change
- [x] Feature

## Screenshots

image

### DIFF
--- a/src/components/PipelineRunDetailsView/visualization/utils/pipelinerun-graph-utils.ts
+++ b/src/components/PipelineRunDetailsView/visualization/utils/pipelinerun-graph-utils.ts
@@ -118,6 +118,29 @@ export const appendStatus = (
     } else if (mTask.status && !mTask.status.reason) {
       mTask.status.reason = RunStatus.Pending;
     }
+
+    // Determine any task test status
+    const taskRunName =
+      pipelineRun.status.taskRuns &&
+      Object.keys(pipelineRun.status.taskRuns).find(
+        (key) => pipelineRun.status.taskRuns[key].pipelineTaskName === mTask.name,
+      );
+    const taskRun = taskRunName && pipelineRun.status.taskRuns[taskRunName];
+    if (taskRun?.status?.taskResults) {
+      const testOutput = taskRun?.status?.taskResults.find(
+        (result) => result.name === 'HACBS_TEST_OUTPUT',
+      );
+      if (testOutput) {
+        try {
+          const outputValues = JSON.parse(testOutput.value);
+          if (outputValues.result === 'FAILURE' || outputValues.result === 'WARNING') {
+            mTask.testStatus = RunStatus.Cancelled;
+          }
+        } catch (e) {
+          // ignore
+        }
+      }
+    }
     return mTask;
   });
 };
@@ -223,6 +246,7 @@ const getGraphDataModel = (pipeline: PipelineKind, pipelineRun?: PipelineRunKind
         height: DEFAULT_NODE_HEIGHT,
         runAfterTasks,
         status: vertex.data.status?.reason,
+        testStatus: vertex.data.testStatus,
         whenStatus: taskWhenStatus(vertex.data),
         task: vertex.data,
       }),
--- a/src/components/PipelineRunDetailsView/visualization/utils/pipelinerun-graph-utils.ts
+++ b/src/components/PipelineRunDetailsView/visualization/utils/pipelinerun-graph-utils.ts
@@ -133,9 +133,8 @@ export const appendStatus = (
       if (testOutput) {
         try {
           const outputValues = JSON.parse(testOutput.value);
-          if (outputValues.result === 'FAILURE' || outputValues.result === 'WARNING') {
-            mTask.testStatus = RunStatus.Cancelled;
-          }
+          mTask.testFailCount = parseInt(outputValues.failures, 10);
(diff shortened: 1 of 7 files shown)


src/components/PipelineRunDetailsView/visualization/utils/pipelinerun-graph-utils.ts
```diff
@@ -118,6 +118,31 @@ export const appendStatus = (
...
+    const taskRun = taskRunName && pipelineRun.status.taskRuns[taskRunName];
+    if (taskRun?.status?.taskResults) {
+      const testOutput = taskRun?.status?.taskResults.find(
+        (result) => result.name === 'HACBS_TEST_OUTPUT',
+      );
+      if (testOutput) {
+        try {
+          const outputValues = JSON.parse(testOutput.value);
+          if (outputValues.result === 'FAILURE') {
+            mTask.testStatus = RunStatus.Failed;
```
karthikjeeyar, line 137: After seeing this in action, showing failed nodes in visualization and success (actual Tekton status) in the logs tab could create confusions (Tekton will stop execution of following tasks if there is a real failure but in this case it may look like it continued execution).  I think Its best we just show a warning to the user if there is a CVE failure/warning and then redirection to logs should help the user to see more details. wdyt?

Screenshot 2023-02-27 at 10 39 20 AM
Screenshot 2023-02-27 at 10 39 41 AM

jeff-phillips-18: I would agree

src/components/PipelineRunDetailsView/visualization/utils/pipelinerun-graph-utils.ts
```diff
@@ -118,6 +118,29 @@ export const appendStatus = (
...
+    const taskRun = taskRunName && pipelineRun.status.taskRuns[taskRunName];
+    if (taskRun?.status?.taskResults) {
+      const testOutput = taskRun?.status?.taskResults.find(
+        (result) => result.name === 'HACBS_TEST_OUTPUT',
+      );
+      if (testOutput) {
+        try {
+          const outputValues = JSON.parse(testOutput.value);
+          if (outputValues.result === 'FAILURE' || outputValues.result === 'WARNING') {
+            mTask.testStatus = RunStatus.Cancelled;
```
simonbaird, line 137: Does it mean that the warnings and failures are visually identical in the UI?

It would be nice if the user could easily see the difference between a warning and a failure, since the outcome is likely to be quite different. E.g. a build could be allowed to be deployed if there's a warning, but would be blocked if there is a failure.

simonbaird: Just noticed the other comments above about this topic.

I still do think user should be able to see visually the difference between a warning and failure in HACBS_TEST_OUTPUT, but having the failed test result look similar to a failed taskrun could lead to confusion.

Could there somehow be two levels of warning icon, e.g. "minor warning" for warnings in HACBS_TEST_OUTPUT, "major warning" for failures?

jeff-phillips-18: Waiting on UX to come up with a way to differentiate these. @andrew-ronaldson

andrew-ronaldson:
Just came from a design share and wanted to share some possible variations on the warning but without adding a new status. Let me know what you think.

1. Red warning is failure and stops progress 
2. Warning with label is test status and; 
3. Warning on a task but doesn't stop the pipeline

Screen Shot 2023-03-01 at 10 52 33 AM

jeff-phillips-18:
@andrew-ronaldson So an test warning just shows the warning status on the node. A test failure shows the warning status along with a `Test failed` badge?

andrew-ronaldson:
Screen Shot 2023-03-01 at 5 08 21 PM
We can refine this more but this should work as a starting point. Use the warning state with counter that shows a count issues.

jeff-phillips-18:
Updated to show badges with counts. Updated the description with the screenshot.

src/components/PipelineRunDetailsView/visualization/utils/pipelinerun-graph-utils.ts
```diff
@@ -118,6 +118,28 @@ export const appendStatus = (
...
+    const taskRun = taskRunName && pipelineRun.status.taskRuns[taskRunName];
+    if (taskRun?.status?.taskResults) {
+      const testOutput = taskRun?.status?.taskResults.find(
+        (result) => result.name === 'HACBS_TEST_OUTPUT',
+      );
+      if (testOutput) {
+        try {
+          const outputValues = JSON.parse(testOutput.value);
+          mTask.testFailCount = outputValues.failures;
+          mTask.testWarnCount = outputValues.warnings;
```
karthikjeeyar, line 137: we need to also check for `result === FAILURE` before setting the testFailCount/testWarnCount because now the failed badge is showing even if the result is SUCCESS 

Screenshot 2023-03-03 at 6 39 13 PM

PR #448, opened 2023-02-28
Don't complete form on editable label enter key

jeff-phillips-18:
## Fixes 
Fixes [HAC-2120](https://issues.redhat.com/browse/HAC-2120)

## Description
Don't submit form on an enter key press in the editable label field. The edit should be saved but the form should not be submitted.

## Type of change
- [x] Bugfix

/cc @karthikjeeyar @christianvogt


jeff-phillips-18:
Closing due to build issue with this PR number

PR #449, opened 2023-02-28
Show `Build starting` for component status when pipeline run does not exist

jeff-phillips-18:
## Fixes 
Fixes [HAC-2344](https://issues.redhat.com/browse/HAC-2344)

## Description
When the Pipelinerun has not been started yet, display the component status as `Build Starting`

## Type of change
- [x] Bugfix

/cc @karthikjeeyar @christianvogt

### DIFF
--- a/src/components/Components/BuildStatusColumn.tsx
+++ b/src/components/Components/BuildStatusColumn.tsx
@@ -1,9 +1,10 @@
 import * as React from 'react';
 import { Button, Flex, FlexItem } from '@patternfly/react-core';
 import { useLatestPipelineRunForComponent } from '../../hooks/usePipelineRunsForApplication';
-import { pipelineRunFilterReducer } from '../../shared';
+import { pipelineRunFilterReducer, runStatus } from '../../shared';
 import { ComponentKind } from '../../types';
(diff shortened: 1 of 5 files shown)


src/components/Components/BuildStatusColumn.tsx
```diff
@@ -49,19 +49,17 @@ const BuildStatusColumn: React.FC<BuildStatusComponentProps> = ({ component, all
   );
 
   const pipelineRun = useLatestPipelineRunForComponent(component);
-  const status = pipelineRunFilterReducer(pipelineRun);
+  const status = pipelineRun ? pipelineRunFilterReducer(pipelineRun) : runStatus.PipelineNotStarted;
```
karthikjeeyar, line 52: when pipelinerun fetch call is inflight during initial component tab load, the `Build Starting` status shows for components that has successfully built earlier. Shall we include if pipelinerun loaded check as well to avoid it ?

jeff-phillips-18: Added check below

karthikjeeyar:
Verified locally works as expected
/approved

karthikjeeyar:
Verified locally works as expected

/approve

PR #451, opened 2023-03-01
remove validation errors when switching back from samples to import  flow

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/HAC-2992

## Description

When switching to samples flow we are resetting the `source.git.url` to empty, but the fieldTouched property is not reset causing the validation message to show up when switching back to import from code flow. 

## Type of change

- [x] Bugfix

Before:
before_fix

After:
fix_validation_issue

## How to test or reproduce?

1. In Add component page, click on the Git url and move to `Start with a Sample` flow
2. Switch back to `Import from your code` flow from samples page.

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge


rottencandy (approved):
/lgtm

PR #462, opened 2023-03-06
add `Dockerfile` option instead of `Other` in runtime selector

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/HAC-3347

https://issues.redhat.com/browse/HAC-3351
https://issues.redhat.com/browse/HAC-3372

## Description

In runtime selector, add `Dockerfile`option instead of `Other`

## Type of change

- [x] Bugfix

https://user-images.githubusercontent.com/9964343/223156485-343cb0f1-1a01-41b4-961a-27641e8e267c.mp4

## How to test or reproduce?

1. Add a repository and in the review step, you can see new Dockerfile option.

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

### DIFF
--- a/src/components/ImportForm/ReviewSection/RuntimeSelector.tsx
+++ b/src/components/ImportForm/ReviewSection/RuntimeSelector.tsx
@@ -41,6 +41,8 @@ export const RuntimeSelector: React.FC<RuntimeSelectorProps> = ({ detectedCompon
         git: { url: sourceUrl, revision, context },
       },
       isDetected,
+      detectionFailed,
+      initialDetectionLoaded,
       components,
     },
     setFieldValue,
@@ -62,10 +64,10 @@ export const RuntimeSelector: React.FC<RuntimeSelectorProps> = ({ detectedCompon
           value: s.name,
           icon: <img src={s.icon.url} />,
         })),
-        dockerFileSample,
+        ...(!detectionFailed ? [dockerFileSample] : []),
       ] || []
     );
-  }, [samples]);
+  }, [samples, detectionFailed]);
 
   const onChange = (value: string) => {
     const runtime = samples.find((s) => s.name === value);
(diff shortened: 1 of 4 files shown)


src/components/ImportForm/ReviewSection/RuntimeSelector.tsx
```diff
@@ -45,22 +46,43 @@ export const RuntimeSelector: React.FC<RuntimeSelectorProps> = ({ detectedCompon
...
   const DetectingRuntime = 'Detecting runtime...';
+  const dockerFileSample = React.useMemo(
+    () => ({
+      key: 'docker-build',
+      value: 'Dockerfile',
+      icon: <DockerIcon color="#2496ed" />,
+      runtime: { name: 'Dockerfile' },
+    }),
+    [],
+  );
```
christianvogt, line 57: This can be a const outside of the component instead of a memo.

src/components/ImportForm/ReviewSection/RuntimeSelector.tsx
```diff
@@ -45,22 +46,43 @@ export const RuntimeSelector: React.FC<RuntimeSelectorProps> = ({ detectedCompon
...
     return (
-      samples?.map((s) => ({ key: s.uid, value: s.name, icon: <img src={s.icon.url} /> })) || []
+      [
+        ...(samples || []).map((s) => ({
+          key: s.uid,
+          value: s.name,
+          icon: <img src={s.icon.url} />,
+        })),
+        dockerFileSample,
+      ] || []
```
christianvogt, line 68: It was discussed that if a dockerfile is detected, we shouldn't remove the dockerfile (previously named 'Other') option. But only if the dockerfile was detected should we provide this option.

This will allow the user to _always_ choose dockerfile as an option. But that's not a viable solution right now without the ability for the user to set the dockerfile location and build context. I believe that without those inputs, there is no way that this option would work if a dockerfile wasn't already detected.

Maybe delivering this right now is ok because we plan to support these new inputs soon.

src/components/ImportForm/ReviewSection/RuntimeSelector.tsx
karthikjeeyar: updated.

rottencandy:
Tested the changes, looks good.

PR #465, opened 2023-03-07
add sidepanel for task run nodes of pipeline run visualization

christianvogt:
## Fixes 
https://issues.redhat.com/browse/HAC-3357

## Description

Adds click interaction for task run nodes of the pipeline run visualization.
Opens a sidepanel with extra information.

The side panel is configured as `inline` which will push the content. The reason for doing this is because it allows us to navigate the visualization and gain access to the right most nodes without having to first close the panel.

When a node is clicked, we scroll it into view to ensure the task node is visible when the panel opens.

Some tasks aren't very interesting, such as skipped tasks. Skipped tasks have no start time or duration or logs.

Known issues:
~~There is a layout issue with consoledot that prevents me from defining two distinct scroll areas: one for the panel and one for the adjacent content. Because of this the panel can scroll off screen. https://issues.redhat.com/browse/RHCLOUD-24606~~
~~This is also causing our logs viewer to scroll the panel down when switching nodes while on the logs tab.~~

Addressed this issue in the second commit. This is a work around to https://issues.redhat.com/browse/RHCLOUD-24606 by setting the main layout to have a `relative` css `position`. By doing this I can then establish a height in the `NamespacedPage` layout. This affects all pages because it establishes a new layout container. We can revisit this once the issue is looked at by the consoledot team.

image

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

image

Success nodes include results:
image

image

Some tasks have no description.
Skipped tasks have no start time or duration or logs.
image
image

## How to test or reproduce?
Create a build such that there is at least one pipeline run.
Navigate to the pipeline run details page.
Select a node to open the side panel.

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/NamespacedPage/NamespacedPage.scss
+++ b/src/components/NamespacedPage/NamespacedPage.scss
@@ -2,4 +2,18 @@
   display: flex;
   flex-direction: column;
   flex: 1 1 auto;
+
+  // TODO work around to https://issues.redhat.com/browse/RHCLOUD-24606
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  right: 0;
+  left: 0;
+
+  .main-layout-container {
+    display: flex;
+    flex-direction: column;
+    flex: 1 1 auto;
+    overflow: auto;
+  }
 }
--- a/src/components/NamespacedPage/NamespacedPage.tsx
+++ b/src/components/NamespacedPage/NamespacedPage.tsx
@@ -14,6 +14,14 @@ type NamespacedPageProps = {
 const NamespacedPage: React.FunctionComponent<NamespacedPageProps> = ({ children }) => {
   const { workspacesLoaded } = React.useContext(WorkspaceContext);
 
+  // TODO work around to https://issues.redhat.com/browse/RHCLOUD-24606
+  React.useEffect(() => {
+    const main = document.querySelector<HTMLElement>('main.pf-c-page__main');
+    if (main) {
+      main.style.position = 'relative';
+    }
+  }, []);
+
   if (!workspacesLoaded) {
     return (
       <Bullseye>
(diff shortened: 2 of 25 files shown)


jeff-phillips-18:
Is it worth allowing the skipped tasks to be selectable? There won't be a description or any more info.

beaumorley:
Would it be possible to put some kind of information for the user if the log tab has no data?  Right now it displays "-" or "No logs found". If we had the reason for each of these cases I can get some short copy to put in. It seems it is either because task is skipped or something else?
This looks great BTW!

christianvogt:
@beaumorley i've updated the messages for the empty tabs.

Skipped tasks currently have no data available to show.

MariaLeonova:
@christianvogt this looks great! Will there be a way to pull in a description for skipped tasks?  Or is that not possible?

jeff-phillips-18:
When I switch to the PLR page, the screen animates growing from left to right:

Kapture 2023-03-10 at 13 44 12

Nodes should have a pointer cursor indicating they can be clicked.

christianvogt:
@jeff-phillips-18 added pointer cursor.

Drawer is a PF issue. I'm still investigating what the cause may be because the drawer starts out closed by default

mattreid:
@christianvogt For a task with no description, do you have a special message or just hide the description field?

christianvogt:
@mattreid the description field is displayed with `-` which is the pattern that's been used elsewhere when there's no data to display.

karthikjeeyar:
@christianvogt Latest height changes is impacting the logs tab content height.

MariaLeonova:
@christianvogt one thing I noticed when interacting with the side panel is that the visualisation doesn't scroll when I move from node to node. So if I select a node on the right of the page, the side panel would cover it. Wonder if there's an easy way to center the selected node?
Screenshot 2023-03-15 at 12 07 26

jeff-phillips-18:
@MariaLeonova It should auto scroll the selected node into view. It is working for me.

MariaLeonova:
https://user-images.githubusercontent.com/14108590/225306807-6ab94606-c995-4c44-8605-cc08468f7334.mov

In firefox this happens^^

jeff-phillips-18:
Seems to be a Firefox thing. Works fine in Chrome, Edge, and Safari

christianvogt:
@MariaLeonova thanks, i'll file a bug for FireFox and see what's going on. I implemented the auto scroll and it's functional on webkit browsers such as Chrome.

PR #470, opened 2023-03-10
Fix for Commit Visualization error

jeff-phillips-18:
## Fixes 
Fixes [HAC-3416](https://issues.redhat.com/browse/HAC-3416)

## Description
Fix crash in commit visualization due to incorrectly set node type

## Type of change
- [x] Bugfix

/cc @karthikjeeyar


karthikjeeyar (approved):
Verified locally, works as expected

/lgtm

PR #489, opened 2023-03-15
periodic test run login

subhashkhileri:
## Fixes 

periodic test run login

## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Build related changes
- [ ] Other (please describe): test

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/integration-tests/support/commands/hooks.ts
+++ b/integration-tests/support/commands/hooks.ts
@@ -17,7 +17,7 @@ before(() => {
     JSON.stringify({ 'application-list-getting-started-modal': true }),
   );
 
-  if (Cypress.env('PR_CHECK') === true) {
+  if (Cypress.env('PR_CHECK') || Cypress.env('PERIODIC_RUN')) {
     Login.prCheckLogin();
   } else {
     Login.login();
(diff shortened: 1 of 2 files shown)


integration-tests/support/commands/hooks.ts
```diff
@@ -17,7 +17,7 @@ before(() => {
     JSON.stringify({ 'application-list-getting-started-modal': true }),
   );
 
-  if (Cypress.env('PR_CHECK') === true) {
+  if (Cypress.env('PR_CHECK') || Cypress.env('PERIODIC_RUN')) {
```
jrichter1, line 20: `Cypress.env('PR_CHECK')` is going to have a default value of `'false'`, which in this case is still going to produce a truthy result.
In this configuration it pretty much always acts as if it's a pr check.

subhashkhileri: when `Cypress.env('PERIODIC_RUN')` is `true` I want to execute this if block for ephemeral env to login

subhashkhileri: Update the `PR_CHECK` to boolean in cofig

PR #495, opened 2023-03-16
Add build config settings for adding components

jeff-phillips-18:
## Fixes 
Fixes [HAC-3350](https://issues.redhat.com/browse/HAC-3350)

## Description
Add build configuration settings when adding components. Allow override of build context directory and dockerfileUrl.

image

/cc @karthikjeeyar @Misjohns


Misjohns:
@jeff-phillips-18  Based on Elson's slack message the Docker file path default would not include the "./" since that is redundant with the preceding Build context directory field default. The placeholder text would be "Dockerfile".

jeff-phillips-18:
What is shown is not placeholder text, that is the value we are retrieving from the ComponentDetectionQuery. When they fix their end, it will fix the UI. If there is no value returned, the placeholder text would show.

christianvogt:
/lgtm cancel

How can we be overriding the same value from the git options `context` on the first page of the wizard?

christianvogt:
It was concluded in slack that for now we should omit the context dir on this page and just provide users with the the dockerfile path field. This is because there is no way to set the build context via a Component right now. This will be re-evaluated in the future if necessary.

Misjohns:
Let's remove the "> Build configuration" section. Change "> Deploy configuration" to "> Build & deploy configuration". In that section we'd have Docker file path and Target port, then CPU and Memory then "> Show advanced...."

jeff-phillips-18:
@Misjohns @christianvogt Updated to include only the Docker file path and require the docker path to be a relative path or a URL.

PR #499, opened 2023-03-17
Add access checks to actions

rottencandy:
## Fixes 
https://issues.redhat.com/browse/HAC-3343

## Description
Add access checks to buttons & actions and disable them if user does not have sufficient access.

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

(No design changes)

## Browser conformance: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


christianvogt:
@beaumorley please comment as to when actions should be hidden vs disabled because a user cannot perform the action due to permissions. Eg. a user visiting another workspace that they do not own.

beaumorley:
We have these general conventions around RBAC: http://openshift.github.io/openshift-origin-design/designs/administrator/future-openshift/role-based-access-control/

Applying this to your example, if the user does not own the workspace but can read it, then the disable actions would be appropriate. If they do not have read access, the workspace can be hidden and we can use an empty state with a No Access message. LMK If you need messaging for any Empty States so they match the existing patterns. I can also work with Missy to provide more detailed guidance on specific screens or flows.

Misjohns:
We could use a No Access empty state: https://www.patternfly.org/v4/components/empty-state/design-guidelines/#no-access

rottencandy:
@beaumorley @Misjohns Access checks for full pages will be done in https://issues.redhat.com/browse/HAC-3430

For this PR, I have disabled buttons and links for all the actions that a user does not access to perform.

src/utils/rbac.ts
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import { k8sCreateResource, K8sModelCommon, K8sVerb } from '@openshift/dynamic-plugin-sdk-utils';
+import memoize from 'lodash/memoize';
+import { SelfSubjectAccessReviewModel } from '../models/rbac';
+import { AccessReviewResourceAttributes, SelfSubjectAccessReviewKind } from '../types/rbac';
+import { useWorkspaceInfo } from './workspace-context-utils';
+
+export const checkAccess = memoize(async (group, resource, subresource, namespace, verb) => {
```
karthikjeeyar, line 8: noticed an issue with `SelfSubjectAccessReview` network calls. In application details page there has to be four different access review calls for the resources like Applications, environments, components, integrationTest but only one call (components) is made and that result is passed back to all other access review caller functions.
 
 Application details page access network call:
 
Screenshot 2023-03-21 at 1 11 01 PM


The issue is with memoize function in the checkAcess method, by default lodash will use the first argument passed as a cache key in this case the first argument is `group` .. since all four resources I mentioned is using the same group `appstudio.redhat.com` so all of them are sharing the same result. A better approach would be pass a second argument to the memoize function to resolve the cache key. something like `<group>-<resource>-<verb>` to be more precise.

karthikjeeyar:
[PipelineRunEmptyState](https://github.com/openshift/hac-dev/blob/main/src/components/PipelineRunDetailsView/PipelineRunEmptyState.tsx#L32)'s button is missing access check reviews.

src/utils/rbac.ts
rottencandy: Thanks, did not realize this. Updated.

rottencandy:
> @rottencandy [IntegrationTestDetailsView](https://github.com/openshift/hac-dev/blob/main/src/components/IntegrationTest/IntegrationTestDetailsView.tsx#L67-L90) actions are missing access checks.

Updated.

beaumorley:
@Misjohns So if the button is disabled, can we show a generic tooltip for now? Can we ask Abi for copy?

Misjohns:
Let's use "You do not have access to add a component" so that it is a little more specific than "perform this action"? So the first part would be the same followed by the <action to be performed> "You do not have access to <action description>"

rottencandy:
Thanks, updated all the button tooltips to be more specific, does this look good:

<details>

<summary>Screenshots</summary>





























</details>

abigaeljamie (commented):
Looks great! Can we replace "do not" with "don't"? It will help us align more with PatternFly's guidelines and adhere a little closer to our friendlier brand voice: https://www.patternfly.org/v4/ux-writing/writing-for-all-audiences

abigaeljamie (approved):
Thank you!

karthikjeeyar:
LGTM! 

/hold cancel

PR #503, opened 2023-03-21
Introducing pact test

Katka92:
## Description
Add pact tests to the hac-dev repo. Those tests will not be executed now, the job for running them on a PR check is in progress here https://github.com/openshift/release/pull/37401. 

Related issue https://issues.redhat.com/browse/HAC-2988


pact-tests/application-creation.pact.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+        },
+      };
+      const expectedResponse = {
+        apiVersion: 'appstudio.redhat.com/v1alpha1',
+        kind: 'Application',
+        metadata: {
+          creationTimestamp: term({
+            generate: '2022-01-21T13:36:30Z',
+            matcher:
+              '^(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9]):([0-5][0-9]):([0-5][0-9])(.[0-9]+)?(Z|[+-](?:2[0-3]|[01][0-9]):[0-5][0-9])?$',
```
jrichter1, line 39: these matchers look reusable, could you extract them to a common place?

I'm assuming this matches a date+time, it would be more clear if it had a nice name. And I wouldn't have to look at that expression so many times.

PR #504, opened 2023-03-21
fix task run step duration in tooltip

christianvogt:
## Fixes 
https://issues.redhat.com/browse/HAC-3528

## Description
All task run steps get the same start time in the task run status even though steps are run sequentially.

Updated the logic to compute the duration of running steps based on the finish time of the previous step in the sequence. Once a step completes, it gets a proper start and finish time.

Also updates the status of a step to be in the pending state until the previous step in the sequence is complete.

Improved the duration display to update every second instead of updating only when the resource gets updated.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

image
image
image

## How to test or reproduce?
Start a new build.
Hover over the `build-container` node in the pipeline run visualization

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


rohitkrai03 (approved):
/lgtm

PR #546, opened 2023-04-04
User provided component name should not be overridden by cdq defaults

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/HAC-3585

## Description

If the user has provided a component name, then it should not be updated when a runtime is changed.

## Type of change

- [x] Bugfix

https://user-images.githubusercontent.com/9964343/229850506-19e7ed42-d994-401c-adf3-326aee724e5e.mov

## How to test or reproduce?

1. In the review step, change the component name
2. choose a different runtime

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge


sahil143 (approved):
/lgtm
verified locally

PR #550, opened 2023-04-05
Update task name for scraping EC report output

simonbaird:
## Fixes 

No Jira filed currently.

## Description

There were some changes to the Enterprise Contract task definition and now the step containing the Enterprise Contract output in YAML format is called "report" instead of "validate".

See the task definition change here:
https://github.com/enterprise-contract/ec-cli/pull/589

This is related to the recently merged #530 .

## Type of change

- Bugfix

## How to test or reproduce?

I believe the newly merged code from #530 won't be working as expected in RHTAP stage and prod due to the recent step name change.

I'm not sure how to test this, or whether there are other changes required. Feel free to advise if there's something else needed. Attention @christianvogt @sahil143 .


simonbaird:
FWIW if this is problematic, or difficult to merge, we could instead switch the step names in the task definition back to what they were before. We wouldn't revert https://github.com/enterprise-contract/ec-cli/pull/589 but we could change the step name for the report step back to validate and call the validate step something else.

PR #564, opened 2023-04-12
Fix width of visualization nodes based on labels

jeff-phillips-18:
## Fixes 
Fixes [RHTAPBUGS-31](https://issues.redhat.com/browse/RHTAPBUGS-31)

## Description
When determining node width, use the displayed label on the node not just the resource name.

## Type of change
- [x] Bugfix

/cc @karthikjeeyar


christianvogt:
Not sure if it's related but i got this with 2 components and 2 integration tests:
image

PR #565, opened 2023-04-12
fix build logs autoscroll performance issue

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/RHTAPBUGS-107

## Description

Autoscrolling is not working for larger logs data stream (eg: build-container task).

**Root Cause Analysis**

The performance bottleneck is caused when a large dataset is streamed, the `scrollIntoView` method is being called multiple times continuously which blocks the javascript thread and the browser becomes unresponsive.

**Solution:**

Throttling `scrollIntoView` method and made the function call asynchronous by wrapping in setTimeout function.

## Type of change

- [x] Bugfix

**Before:**

https://user-images.githubusercontent.com/9964343/231555950-cc55a47f-3bff-484a-9731-3e6104363c22.mov

**After:**

https://user-images.githubusercontent.com/9964343/231555969-92527e72-00a0-4415-8844-906c545916d5.mov

## How to test or reproduce?

1. Add a quarkus or spring boot application from samples.
2. view build logs (build-container task)

## Unit tests

```
Logs
    ✓ should not render logs component if the render is set to false (48 ms)
    ✓ should render logs component (15 ms)
    ✓ should show the error alert (13 ms)
    ✓ should fetch data from web sockets (5 ms)
```    

```
  MultiStreamLogs
    ✓ should render loading indicator (38 ms)
    ✓ should render the logs control buttons and container (17 ms)
    ✓ should render resume log stream button when user scrolls up and removed when users scrolls to the bottom of the logs (31 ms)
    ✓ should call onDownload callback when download button is pressed (12 ms)
    ✓ should handle if the download fails (28 ms)
 ```  

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

cc: @mattreid @rohitkrai03

### DIFF
--- a/src/shared/components/pipeline-run-logs/logs/Logs.tsx
+++ b/src/shared/components/pipeline-run-logs/logs/Logs.tsx
@@ -8,6 +8,7 @@ import {
 } from '@openshift/dynamic-plugin-sdk-utils';
 import { Alert } from '@patternfly/react-core';
 import { Base64 } from 'js-base64';
+import { throttle } from 'lodash-es';
 import { PodModel } from '../../../../models/pod';
 import { ContainerSpec, PodKind } from '../../types';
 import { LOG_SOURCE_TERMINATED } from '../utils';
@@ -46,16 +47,24 @@ const Logs: React.FC<LogsProps> = ({
 
   const appendMessage = React.useRef<(blockContent) => void>();
 
+  const safeScroll = throttle(
+    () =>
+      requestAnimationFrame(() => {
+        scrollToRef.current.scrollIntoView({ behavior: 'smooth' });
+      }),
+    300,
+  );
+
   appendMessage.current = React.useCallback(
     (blockContent: string) => {
       if (contentRef.current && blockContent) {
         contentRef.current.innerText += blockContent;
       }
       if (scrollToRef.current && blockContent && render && autoScroll) {
-        scrollToRef.current.scrollIntoView({ behavior: 'smooth' });
+        safeScroll();
       }
     },
-    [autoScroll, render],
+    [autoScroll, render, safeScroll],
   );
 
   if (resourceStatusRef.current !== resourceStatus) {
@@ -114,9 +123,9 @@ const Logs: React.FC<LogsProps> = ({
 
   React.useEffect(() => {
     if (scrollToRef.current && render && autoScroll) {
-      scrollToRef.current.scrollIntoView({ behavior: 'smooth' });
+      safeScroll();
     }
-  }, [autoScroll, render]);
+  }, [autoScroll, render, safeScroll]);
 
   return (
     <div className="logs" style={{ display: render ? '' : 'none' }}>
--- a/src/shared/components/pipeline-run-logs/logs/Logs.tsx
+++ b/src/shared/components/pipeline-run-logs/logs/Logs.tsx
@@ -8,7 +8,6 @@ import {
 } from '@openshift/dynamic-plugin-sdk-utils';
 import { Alert } from '@patternfly/react-core';
 import { Base64 } from 'js-base64';
-import { throttle } from 'lodash-es';
 import { PodModel } from '../../../../models/pod';
 import { ContainerSpec, PodKind } from '../../types';
 import { LOG_SOURCE_TERMINATED } from '../utils';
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/shared/components/pipeline-run-logs/logs/Logs.tsx [hunk 1]
```diff
@@ -46,16 +47,22 @@ const Logs: React.FC<LogsProps> = ({
 
   const appendMessage = React.useRef<(blockContent) => void>();
 
+  const safeScroll = throttle(() =>
+    setTimeout(() => {
```
christianvogt, line 51: why is this done in a timeout; is it even necessary? If so why not `requestAnimtationFrame` ?

btw it looks like ODC throttles updates every 300ms.

karthikjeeyar: I wrapped in setTimeout to make the call async and to make the scroll experience smoother, now switched to `requestAnimationFrame` and added 300 ms throttling as well.

christianvogt:
I'm having issues where after scrolling up then all the way back down, the auto scroll feature won't work.

karthikjeeyar:
> I'm having issues where after scrolling up then all the way back down, the auto scroll feature won't work.

Fixed it.

mattreid:
This could also complicate things, but would it help if the tail functionality was toggled or enabled somehow? If I scroll up and go looking for something, I probably don't want to fight it dragging me back down, but it would be good to have a way to get it synced up again when I'm done looking, either by turning something back on, or by scrolling back to the bottom.

karthikjeeyar:
@mattreid @christianvogt I have updated the code to show a `Resume log stream` button at the bottom of the logs container, which will be visible when the user scrolls up to view some content. let me know if this solution is fine 


https://user-images.githubusercontent.com/9964343/231914035-f911274a-b8da-4c94-a5ed-ec73d0c2c960.mp4

Misjohns:
Can we add the resume icon to the footer?
https://www.patternfly.org/v4/extensions/log-viewer/react-demos#with-complex-toolbar

image

mattreid:
That looks good to me, thank you @karthikjeeyar!

src/shared/components/pipeline-run-logs/logs/Logs.tsx
```diff
@@ -41,21 +42,37 @@ const Logs: React.FC<LogsProps> = ({
...
+    blockContentRef.current = '';
+  }, 200);
+
+  const safeScroll = throttle(
+    () =>
+      requestAnimationFrame(() => {
+        scrollToRef.current.scrollIntoView({ behavior: 'smooth' });
+      }),
+    100,
+  );
```
christianvogt, line 63: This should be wrapped as a `React.useCallback`

src/shared/components/pipeline-run-logs/logs/Logs.tsx
```diff
@@ -41,21 +42,37 @@ const Logs: React.FC<LogsProps> = ({
...
+
   const onCompleteRef = React.useRef<(name) => void>();
   onCompleteRef.current = onComplete;
 
   const appendMessage = React.useRef<(blockContent) => void>();
 
+  const addContent = debounce(() => {
+    contentRef.current.innerText += blockContentRef.current;
+    blockContentRef.current = '';
+  }, 200);
```
christianvogt, line 55: This should be wrapped as a `React.useCallback`. However once that's done, i suspect you can run into issues where updates continue to happen within the 200ms and therefore the screen never gets updated until we get a large chunk. I suggest throttling instead of debouncing here.

src/shared/components/pipeline-run-logs/logs/Logs.tsx
```diff
@@ -41,21 +42,37 @@ const Logs: React.FC<LogsProps> = ({
...
   appendMessage.current = React.useCallback(
     (blockContent: string) => {
       if (contentRef.current && blockContent) {
-        contentRef.current.innerText += blockContent;
+        blockContentRef.current += blockContent;
+        addContent();
       }
       if (scrollToRef.current && blockContent && render && autoScroll) {
-        scrollToRef.current.scrollIntoView({ behavior: 'smooth' });
+        safeScroll();
```
christianvogt, line 72: Why is adding content and scrolling in separate debounce / throttle ?

In ODC we do this:
```
  const addContentAndScroll = React.useCallback(
    throttle(() => {
      if (contentRef.current) {
        contentRef.current.innerText += blockContentRef.current;
      }
      if (scrollToRef.current) {
        scrollToRef.current.scrollIntoView({ behavior: 'smooth', block: 'end' });
      }
      blockContentRef.current = '';
    }, 300),
    [],
  );
```

src/shared/components/pipeline-run-logs/logs/Logs.tsx
karthikjeeyar: Throttling dint work, it is causing too many scrollIntoView calls and browser couldn't handle that. However debounce may take very little extra time to update the screen but eventually when it does, there will be one scroll operation that takes us to the bottom of the page, users can scroll up to view content without any issue and also resume the stream back.

src/shared/components/pipeline-run-logs/logs/Logs.tsx
karthikjeeyar: > Why is adding content and scrolling in separate debounce / throttle ?



This issue ([RHTAPBUGS-107](https://issues.redhat.com/browse/RHTAPBUGS-107)) exists in ODC as well, we need fix it there too.  :) 

Adding content (using `elem.innerText`) and `elem.scrollingIntoView` in same throttle call is causing too many blocking operations causing the autoscroll feature to not work and browser goes unresponsive at moments. My Idea was to limit the `scrollIntoView` call especially during the phase where the data moves rapidly, hence used debounce to append the content in batches and perform scroll operation once in 100ms using throttle.

karthikjeeyar:
Thanks Christian for the suggestion, I have removed debounce/throttle functions and now conditionally rescheduling the requestAnimationFrame callback.

PR #568, opened 2023-04-14
e2e: delete applications using api

jrichter1:
## Fixes 
https://issues.redhat.com/browse/HAC-3706

## Description
make after hooks faster and not prone to clicking errors by using the API instead or UI


integration-tests/utils/Applications.ts
```diff
@@ -17,17 +17,27 @@ import { ComponentPage } from '../support/pages/ComponentsPage';
 import { CreateApplicationPage } from '../support/pages/CreateApplicationPage';
 import { ComponentsTabPage } from '../support/pages/tabs/ComponentsTabPage';
 import { OverviewTabPage } from '../support/pages/tabs/OverviewTabPage';
-import { PipelinerunsTabPage } from '../support/pages/tabs/PipelinerunsTabPage';
 import { Common } from './Common';
 
 export class Applications {
   static deleteApplication(applicationName: string) {
```
subhashkhileri, line 23: I think `create-application-from-sample.spec` also uses the same method in test to delete the application.
I think it would be nice to have one test that deletes from UI in it() block.

jrichter1: does it look better now? :)

PR #562, opened 2023-04-11
Add environments nav item and update environments list page

rohitkrai03:
## Fixes 

- https://issues.redhat.com/browse/HAC-3690
- https://issues.redhat.com/browse/HAC-3598

## Description

- Added new navigation item for Environments.
- Updated the logic for workspaced pages.
  - The applications pages will be default and use routes starting with `workspaces`.
  - All the other categories of routes will start with category followed by workspaced context. 
- Refactored and updated all the environments related pages.
- Removed old hacbs related code for environments.

## Type of change

- [x] Feature
- [x] Refactoring 

## Screen shots / Gifs for design review 

https://user-images.githubusercontent.com/6041994/231233852-b924e2a8-2bda-412b-8938-671575f6e2e8.mov

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

PR #571, opened 2023-04-14
[WIP] Add connection status to environment card

rottencandy:
## Fixes 
https://issues.redhat.com/browse/HAC-3599

## Description
Add connection status to environment card using GitOpsDeploymentManagedEnvironment and update card description.

Depends on #562 , only consider commit c816dd44b2ca8b8d367769e205645c90ab501926

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

## Browser conformance: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


pankajshivpuje:
@rottencandy We need to update the environment card to show cluster type and remove application status and last deploy. Only label we are using is for type values. All the other text should be normal (no labels).

rottencandy:
Removed labels and app status fields.

Since the CR doesn't yet provide the cluster type, I've added a hardcoded OpenShift type for now. @rohitkrai03 is this fine?


I've used the default color for the alert message at the top, since it's more consistent with existing alerts in UI, but I can change it if needed.

rohitkrai03:
@rottencandy I don't think we should remove application status from environments card in application details page yet. It's still useful information I think. @pankajshivpuje WDYT?

pankajshivpuje:
> @rottencandy I don't think we should remove application status from environments card in application details page yet. It's still useful information I think. @pankajshivpuje WDYT?

My context was more of environment cards in the environment tab. If there are multiple application deployed on that environment, what application the status points to? I am not aware of how application status is used on application details page. @Ranelim Any thoughts?

rohitkrai03:
> > @rottencandy I don't think we should remove application status from environments card in application details page yet. It's still useful information I think. @pankajshivpuje WDYT?
> 
> My context was more of environment cards in the environment tab. If there are multiple application deployed on that environment, what application the status points to? I am not aware of how application status is used on application details page. @Ranelim Any thoughts?

@pankajshivpuje We don't show application status on environments list page. We only show application status if you're in the context of application details view.

rottencandy:
Environments page:

Application details page:

src/components/Environment/EnvironmentCard.tsx
```diff
@@ -94,15 +103,32 @@ const EnvironmentCard: React.FC<EnvironmentCardProps> = ({ environment, readOnly
...
             <DescriptionListDescription>
-              <Label>{getEnvironmentDeploymentStrategyLabel(environment)}</Label>
+              {getEnvironmentDeploymentStrategyLabel(environment)}
             </DescriptionListDescription>
           </DescriptionListGroup>
           <ApplicationEnvironmentStatus
             environment={environment as EnvironmentKindWithHealthStatus}
           />
+
+          <DescriptionListGroup>
```
rohitkrai03, line 117: This can go before the application status. Having application status info at the end keeps it consistent for both views.

rohitkrai03 (changes requested):

rohitkrai03 (approved):
/lgtm

PR #573, opened 2023-04-17
increase bonfire deploy timeout for pr checks

jrichter1:
bonfire likes to slow down on occasion, trying to increase the timeout so it can still finish


Katka92:
IDK if we need that, they currently had some issues with nodes, they were running out of CPU so a lot of namespaces were not provisioned appropriately.

PR #592, opened 2023-04-28
Redesigned and refactored new import flow

rohitkrai03:
## Fixes 

- https://issues.redhat.com/browse/HAC-3660

## Description

- Updated the import flow to new design.
- Refactored the components based on new flow and removed FormikWizard.

## Type of change

- [x] Feature

## Screen shots / Gifs for design review 

https://user-images.githubusercontent.com/6041994/235164417-ce0b37c9-9fc7-4eca-abc6-e7494807f767.mov

## TODO
- [ ] Fix e2e tests.

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

PR #595, opened 2023-04-28
add build secret section in the import flow

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/HAC-3481

## Description

 As a user, I want to add my token to the namespace so that I can use it in my build pipeinerun's partner tasks (eg: snyk). 
 
 PR contains following changes
 
 1. Supports only `Opaque` secrets at the moment.
 2. List the `snyk-secret`  in the select box, when the user selects the task a predefined token key will be presented to the user and they can enter the token data.
 3. Users can also enter the custom secret name and add one or more key/value pairs.
 4. Drag and drop or Upload a file in value section.
 

     

## Type of change

- [x] Feature

Screenshot 2023-04-29 at 12 34 54 AM

Screenshot 2023-04-29 at 12 34 32 AM

https://user-images.githubusercontent.com/9964343/235234038-8c57586f-7e09-4902-b2e0-655e312ebf1a.mov

snyk_task

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


src/components/ImportForm/ImportForm.tsx
```diff
@@ -36,6 +36,8 @@ const ImportForm: React.FunctionComponent<ImportFormProps> = ({ applicationName
     },
     namespace,
     secret: '',
+    secrets: [],
```
rohitkrai03, line 39: Probably need to name this a bit better to distinguish between auth secret and these secrets.

src/components/ImportForm/SecretSection/SecretSection.tsx
```diff
@@ -0,0 +1,74 @@
...
+        onChange={(v) =>
+          setFieldValue(
+            'secrets',
+            values.secrets.filter((vs) => v.includes(vs.secretName)),
+          )
+        }
+      >
+        {(props) => <InputField name={props.name} type={TextInputTypes.text} isReadOnly />}
+      </TextColumnField>
+      <Flex direction={{ default: 'row' }} justifyContent={{ default: 'justifyContentFlexStart' }}>
```
rohitkrai03, line 55: Do we have a way to show the already created secrets so that user doesn't create the same secrets again?

rohitkrai03 (changes requested):

Misjohns:
For the secret section:

- Secret remove is too far away from the field; use the same placement as the environment variable field placement.
- Secret fields are very long. Let's align those with the variable name field.

Secret bug

src/components/ImportForm/SecretSection/SecretSection.tsx
karthikjeeyar: I thought about this, but the problem is the list will be long if we show all the secrets (even if we filter by type - Opaque). some secrets are already created for the user and some are created by the users. It will show an alert message when you submit a secret with existing name and also I am checking if `snyk-secret` is already created or not, if its created I am not showing that secret in the predefined list.

karthikjeeyar:
@Misjohns  I have fixed the width of the secret field and remove button placement. 

image

src/components/ImportForm/ImportForm.tsx
karthikjeeyar: updated it to `importSecrets`

karthikjeeyar:
I rebased this PR with the new Import flow changes https://github.com/openshift/hac-dev/pull/592

rohitkrai03 (approved):
/lgtm

PR #596, opened 2023-05-02
Allow component instances to be scaled to 0

jeff-phillips-18:
## Fixes 
Addresses [RHTAPBUGS-12](https://issues.redhat.com/browse/RHTAPBUGS-12)

## Description
Fixes the UI side for setting and showing instances that are scaled to zero. However, the backend is clearing the replicas field when it is set to zero so the UI will still not show the scaling to zero.

## Type of change
- [x] Bugfix

/cc @karthikjeeyar @rohitkrai03


christianvogt:
@jeff-phillips-18 when i go to edit the component settings, a value of `0` doesn't show up:

image

PR #600, opened 2023-05-04
Restrict resource name entries to 63 characters

jeff-phillips-18:
## Fixes 
Fixes [RHTAPBUGS-215](https://issues.redhat.com/browse/RHTAPBUGS-215)

## Description
Limits user entry of resource names to 64 characters. Also fixes a text overflow issue for environment cards when the environment name is very long (env names are not restricted in length).

## Type of change
- [x] Bugfix

/cc @christianvogt @rohitkrai03

### DIFF
--- a/src/components/Environment/EnvironmentCard.tsx
+++ b/src/components/Environment/EnvironmentCard.tsx
@@ -7,7 +7,6 @@ import {
   CardActions,
   CardBody,
   Text,
-  TextContent,
   TextVariants,
   Label,
   DescriptionList,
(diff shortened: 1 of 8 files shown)


christianvogt:
Need to restrict env name creation according to https://issues.redhat.com/browse/GITOPSRVCE-573

Not sure adding validation to the env targetNamespace is required. I'd omit that from this PR.

jeff-phillips-18:
@christianvogt Updated to not restrict the target namespace field length. Added the resource name restrictions on environment names.

PR #602, opened 2023-05-05
Fix list view column alignments

jeff-phillips-18:
## Fixes 
Fixes [RHTAPBUGS-206](https://issues.redhat.com/browse/RHTAPBUGS-206)

## Description
Fixes the column alignment for list view caused by the kebab column. PF table removes padding for cells with no content.
Use a space for content for the action header cell and add padding to equal the width of the kebab button.

## Type of change
- [x] Bugfix

/cc @christianvogt @karthikjeeyar


karthikjeeyar:
Taskrun list is misaligned

Screenshot 2023-05-05 at 8 49 13 PM

jeff-phillips-18:
@karthikjeeyar Fixed:

image

PR #613, opened 2023-05-09
recommend better application name based on the component url

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/RHTAPBUGS-237

## Description

Use repo name from component url as a default application name, and allow user to change it in the review step, if it is desired.

## Type of change

- [x] Bugfix

https://user-images.githubusercontent.com/9964343/237036607-20b76e49-c5a9-43c3-9253-25f10ebfe211.mov

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

### DIFF
--- a/src/components/ImportForm/utils/useValidApplicationName.ts
+++ b/src/components/ImportForm/utils/useValidApplicationName.ts
@@ -1,32 +1,44 @@
 import * as React from 'react';
-import { useK8sWatchResource } from '@openshift/dynamic-plugin-sdk-utils';
-import { ApplicationGroupVersionKind } from '../../../models';
+import gitUrlParse from 'git-url-parse';
+import { useApplications } from '../../../hooks/useApplications';
 import { ApplicationKind } from '../../../types';
 import { useWorkspaceInfo } from '../../../utils/workspace-context-utils';
 
 const BASE_NAME = 'my-app';
 
-export const useValidApplicationName = (): [string, boolean, unknown] => {
+export const incrementNameCount = (
+  applications: ApplicationKind[],
+  prefferedName?: string,
+): string => {
+  let name = prefferedName || BASE_NAME,
+    i = 1;
+  while (applications.some((app) => app.metadata.name === name)) {
+    name = `${prefferedName || BASE_NAME}-${i}`;
+    i++;
+  }
+  return name;
+};
+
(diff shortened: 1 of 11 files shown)


karthikjeeyar:
cc: @rohitkrai03 @christianvogt

jeff-phillips-18 (approved):
/lgtm

src/components/ImportForm/SourceSection/SourceSection.tsx
```diff
@@ -91,7 +94,9 @@ const SourceSection: React.FC<SourceSectionProps> = () => {
     if (accessCheckLoaded) {
       if (isRepoAccessible) {
         try {
-          const { organization } = gitUrlParse(sourceUrl);
+          const { organization, name } = gitUrlParse(sourceUrl);
+          !inAppContext && setFieldValue('application', incrementNameCount(applications, name));
```
rohitkrai03, line 98: Do we actually need to do this in SourceSection? Maybe just update the application name in `ApplicationSection`. The hook `useApplicationName` can take a Git URL and return back a valid application based on git repo.

src/components/ImportForm/utils/useValidApplicationName.ts
```diff
@@ -1,31 +1,33 @@
...
+  let name = prefferedName || BASE_NAME,
+    i = 1;
+  while (applications.some((app) => app.metadata.name === name)) {
+    name = `${prefferedName || BASE_NAME}-${i}`;
+    i++;
+  }
+  return name;
+};
+
 export const useValidApplicationName = (): [string, boolean, unknown] => {
```
rohitkrai03, line 21: You would also need to remove the use of this hook from `ImportForm` where it is sending `recommendedApplicationName`.

rohitkrai03 (changes requested):

src/components/ImportForm/SourceSection/SourceSection.tsx
rottencandy: I think we can also update it for sample import flow this way

karthikjeeyar: Moved the hook usage to review section. For samples flow, I am retaining the same recommendation logic as before.

rohitkrai03 (approved):
/lgtm

PR #642, opened 2023-05-18
Don't remove app if test failed.

Katka92:
Don't remove app if the test fails. Applies to Basic Happy Path and Advanced Happy Path tests.
Related issue: [HAC-3982](https://issues.redhat.com/browse/HAC-3982)

### DIFF
--- a/integration-tests/tests/basic-happy-path.spec.ts
+++ b/integration-tests/tests/basic-happy-path.spec.ts
@@ -19,7 +19,18 @@ describe('Basic Happy Path', { tags: ['@PR-check', '@publicRepo'] }, () => {
   const quarkusDeplomentBody = 'Congratulations, you have created a new Quarkus cloud application';
 
   after(function () {
-    Applications.deleteApplication(applicationName);
+    // If some test failed, don't remove the app
+    let allTestsSucceeded = true;
+    if (
+      this.test.parent.eachTest((test) => {
+        if (test.state == 'failed') {
+          allTestsSucceeded = false;
+        }
+      })
+    )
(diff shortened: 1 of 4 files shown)


pr_check.sh
```diff
@@ -68,7 +68,8 @@ COMMON_SETUP="-v $WORKSPACE/artifacts:/tmp/artifacts:Z \
     -e CYPRESS_USERNAME=`echo ${B64_USER} | base64 -d` \
     -e CYPRESS_PASSWORD=`echo ${B64_PASS} | base64 -d` \
     -e CYPRESS_GH_PR_TITLE=${PR_TITLE} \
-    -e GH_COMMENTBODY=${GH_COMMENTBODY}"
+    -e GH_COMMENTBODY=${GH_COMMENTBODY} \
+    -e CYPRESS_REMOVE_APP_ON_FAIL=false"
```
jrichter1, line 72: I'd rather see this as a default value in the config file

integration-tests/tests/basic-happy-path.spec.ts
```diff
@@ -19,7 +19,18 @@ describe('Basic Happy Path', { tags: ['@PR-check', '@publicRepo'] }, () => {
...
-    Applications.deleteApplication(applicationName);
+    // If some test failed, don't remove the app
+    let allTestsSucceeded = true;
+    if (
+      this.test.parent.eachTest((test) => {
+        if (test.state == 'failed') {
+          allTestsSucceeded = false;
+        }
+      })
+    )
```
jrichter1, line 30: am I seeing this correctly, that the whole block is an if condition?

Katka92: :woman_facepalming:

PR #647, opened 2023-05-19
Make import form footer sticky

rottencandy:
## Fixes 
https://issues.redhat.com/browse/RHTAPBUGS-250

## Description
Make import form footer sticky

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

https://github.com/openshift/hac-dev/assets/20013884/9c5a1843-f96e-48c2-a014-97d526c3cc35

## Browser conformance: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


Misjohns:
The footer should go to the edge of the Review & Configure container.
image

christianvogt:
Also the drop shadow extends on the right and left of the footer where it would be best to only extend up.

src/components/ImportForm/GitImportActions.tsx
```diff
@@ -19,6 +33,7 @@ const GitImportActions: React.FunctionComponent<GitImportActionsProps> = ({
     isSubmitting,
     isValidating,
     setErrors,
+    handleReset,
```
rohitkrai03, line 36: You can take the `handleSubmit` method from formikProps as well. No need to pass it down from parent.

rohitkrai03 (changes requested):

src/components/ImportForm/GitImportActions.tsx
rottencandy: Updated.

rohitkrai03 (approved):
/lgtm

PR #574, opened 2023-04-17
Integration test git resolver support

abhinandan13jan:
## Fixes 

https://issues.redhat.com/browse/HAC-3243
https://issues.redhat.com/browse/HAC-3240
https://issues.redhat.com/browse/HAC-3241

## Description

- Updates type of IntegrationTestScenario Kind to ResolveRef
- Adds util functions for parameter and Resolver support
- Updates DetailsPage for ITS
- Updates Create/Edit form of IntegrationTestDetails Page to support git resolver
- Updates List View for Integration Test
- Updates and adds unit test suites

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

### create/edit form
------------------------
Screenshot 2023-04-27 at 5 25 03 PM
Screenshot 2023-04-27 at 5 24 34 PM

---------------------------------
### List view using mock data
------------------------------
Screenshot 2023-04-27 at 6 36 11 PM

-----------------------------------
### Details view and linked page using mock data
------------------------------------
Screenshot 2023-04-27 at 6 14 44 PM
Screenshot 2023-04-27 at 6 15 04 PM

-----------------------
### API response and k8sResource creation(currently failing)

Screenshot 2023-04-27 at 5 25 41 PM
Screenshot 2023-04-27 at 5 27 06 PM
----------------------------

## How to test or reproduce?

Updated unit tests

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

PR #648, opened 2023-05-19
Create EC Integration when creating an Application

lcarva:
## Description

When an IntegrationTestScenario running the Enterprise Contract executes, the UI for the corresponding Application is populated with various security results. This is valuable information to users.

This change makes it so when an Application is created, the EC IntegrationTestScenario is automatically created with default values. This lowers the entrybarrier to accessing the security results.

In the future, this auto creation may move to another RHTAP component, e.g. a new EC controller. Or, it may be desireable to introduce UI controls to customize its initial creation, e.g. different Tekton bundle. It's unclear at this point what our users will prefer.  For this reason, we make the most minimal change possible to introduce the feature to users without investing too much in a certain solution right now.

https://issues.redhat.com/browse/HACBS-2191

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

## How to test or reproduce?

Simply create a new application from a sample or other source. The `IntegrationTestScenario` is created automatically.

### DIFF
--- a/src/components/ImportForm/utils/__tests__/submit-utils.spec.ts
+++ b/src/components/ImportForm/utils/__tests__/submit-utils.spec.ts
@@ -1,4 +1,5 @@
 import { createApplication, createComponent } from '../../../../utils/create-utils';
+import { createIntegrationTest } from '../../../IntegrationTest/IntegrationTestForm/utils/create-utils';
 import { detectComponents } from '../cdq-utils';
 import { createResources, checkApplicationName } from '../submit-utils';
 import { ImportFormValues, ImportStrategy } from './../types';
@@ -11,6 +12,10 @@ jest.mock('../../../../utils/create-utils', () => ({
   createComponent: jest.fn(),
 }));
 
+jest.mock('../../../IntegrationTest/IntegrationTestForm/utils/create-utils', () => ({
+  createIntegrationTest: jest.fn(),
+}));
+
 jest.mock('../cdq-utils', () => ({
   detectComponents: jest.fn(),
 }));
@@ -18,6 +23,7 @@ jest.mock('../cdq-utils', () => ({
 const createApplicationMock = createApplication as jest.Mock;
 const createComponentMock = createComponent as jest.Mock;
 const detectComponentsMock = detectComponents as jest.Mock;
+const createIntegrationTestMock = createIntegrationTest as jest.Mock;
 
 describe('Submit Utils: createResources', () => {
   it('should create application and components', async () => {
@@ -46,6 +52,7 @@ describe('Submit Utils: createResources', () => {
       'test-ws',
     );
     expect(createApplicationMock).toHaveBeenCalledTimes(2);
+    expect(createIntegrationTestMock).toHaveBeenCalledTimes(2);
     expect(createComponentMock).toHaveBeenCalledTimes(2);
   });
 
@@ -74,6 +81,7 @@ describe('Submit Utils: createResources', () => {
       'test-ws',
     );
     expect(createApplicationMock).toHaveBeenCalledTimes(0);
+    expect(createIntegrationTestMock).toHaveBeenCalledTimes(0);
     expect(createComponentMock).toHaveBeenCalledTimes(2);
   });
 
@@ -105,6 +113,7 @@ describe('Submit Utils: createResources', () => {
     ).rejects.toThrow();
     expect(createApplicationMock).toHaveBeenLastCalledWith('test-app', 'test-ns', true);
     expect(createComponentMock).toHaveBeenCalledTimes(0);
+    expect(createIntegrationTestMock).toHaveBeenCalledTimes(0);
   });
 
   it('should not create any resources if component dry run fails', async () => {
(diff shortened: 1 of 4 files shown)


lcarva:
This should **not** be merged during the code freeze.

src/components/ImportForm/utils/__tests__/submit-utils.spec.ts
```diff
@@ -73,6 +80,7 @@ describe('Submit Utils: createResources', () => {
     );
     expect(createApplicationMock).toHaveBeenCalledTimes(0);
     expect(createComponentMock).toHaveBeenCalledTimes(2);
+    expect(createIntegrationTestMock).toHaveBeenCalledTimes(1);
```
simonbaird, line 83: I'm wondering why is it called if createApplicationMock was not called?

simonbaird: ...I guess createResources is after the application creation?

src/components/ImportForm/utils/submit-utils.ts
```diff
@@ -99,6 +101,14 @@ export const createResources = async (formValues: ImportFormValues, strategy: Im
 
   await createSecrets(importSecrets, namespace, false);
 
+  const integrationTestValues: IntegrationTestFormValues = {
+    name: `${applicationName}-enterprise-contract`,
+    pipeline: 'enterprise-contract',
+    bundle: 'quay.io/redhat-appstudio-tekton-catalog/pipeline-enterprise-contract:devel',
```
simonbaird, line 107: Idea: Maybe the default bundle url and the pipeline name could be placed in separate constants somewhere.

simonbaird (approved):

src/components/ImportForm/utils/__tests__/submit-utils.spec.ts
lcarva: `createResources` does create the Application, but only if it's not already created. I need to verify what happens if I add a Component to an existing Application. This might need a small change to only create the IntegrationTestScenario if not already there.

src/components/ImportForm/utils/submit-utils.ts
lcarva: +1 I'd love that. I'm just not familiar enough with this code base to tell where is a good place for it.

src/components/ImportForm/utils/__tests__/submit-utils.spec.ts
lcarva: Yup, that was a bug that manifested itself when adding a component to an existing application. Fixed it by only executing `createIntegrationTest` when an application is created. I also added a dry-run call so if there are any issues `createResources` fails early without creating partial resources.

src/components/ImportForm/utils/submit-utils.ts
```diff
@@ -68,8 +70,16 @@ export const createResources = async (formValues: ImportFormValues, strategy: Im
     componentAnnotations = { [SAMPLE_ANNOTATION]: 'true' };
   }
 
+  const integrationTestValues: IntegrationTestFormValues = {
+    name: `${applicationName}-enterprise-contract`,
+    pipeline: 'enterprise-contract',
+    bundle: 'quay.io/redhat-appstudio-tekton-catalog/pipeline-enterprise-contract:devel',
+    optional: true,
```
simonbaird, line 77: Can we get a spec.params in here I wonder? No idea if this might work, but...

```suggestion
    optional: true,
    spec: {
      params: [
        { 
          name: 'POLICY_CONFIGURATION',
          value: 'github.com/enterprise-contract/config//slsa3',
        },  
      ]
    }
```

simonbaird, line 77: As discussed:

```suggestion
    optional: false,
```

src/components/ImportForm/utils/submit-utils.ts
karthikjeeyar: @lcarva  You could create a const file in integrationTest utils folder ([IntegrationTest/IntegrationTestForm/utils](https://github.com/openshift/hac-dev/tree/main/src/components/IntegrationTest/IntegrationTestForm/utils)) and place these constants there  and you can also move this `integrationTestValues` const to that new location and import this here in submit-utils.

src/components/ImportForm/utils/submit-utils.ts
lcarva: The UI support is not quite there yet for this. I think we should do it as follow up?

lcarva: Yeah... So there's also UI to edit an existing IntegrationTestScenario which doesn't support params yet. We should wait for that. In the mean time, I'll update the default ECP.

simonbaird:
PR has 37 commits now for some reason. Do you want to rebase and clean it up?

src/components/ImportForm/utils/submit-utils.ts
lcarva: Done. Thanks!

lcarva:
> PR has 37 commits now for some reason. Do you want to rebase and clean it up?

Thanks! Had to switch the target branch to `main` from `main-next`.

simonbaird:
I have a feeling this might logically conflict with the git resolver changes in #574 IIUC. Does anyone know the likely timeline for merging #574? If it's soon then maybe this should be held back and rebased on top of that one. WDYT?

lcarva:
> @lcarva Integration git resolver PR #685 is merged and deployed in `console.dev.redhat` environment , Please update this PR to match the new payload spec available in `v1beta1` API. cc: @simonbaird

Thank you! I updated the patched and was able to verify against staging :pray:

simonbaird (approved):

PR #667, opened 2023-05-26
Fix for showing git repo status on change

jeff-phillips-18:
## Fixes 
Fixes [RHTAPBUGS-317](https://issues.redhat.com/browse/RHTAPBUGS-317)

## Description
When the user changes the git repo name, we need to set the validating flag after the sourceURL is updated in order for it to be validated correctly. Also, remove the created `SPIAccessCheckModel` resource when we no longer need it.

## Type of change
- [x] Bugfix


jeff-phillips-18:
Closing in lieu of #675

PR #677, opened 2023-05-31
fix dockerfile icon size in runtime selector dropdown

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/RHTAPBUGS-261

## Description

we use a PF Docker icon in runtime selector, and that has to match the size of other icons (24px)

Note: Other smaller icons `Node and Go` is fetched dynamically from  https://registry.devfile.io/index/sample and it needs to be fixed from HAS team. cc: @elsony 

Before:
Screen Shot 2023-05-11 at 10 57 18 AM

After:

Screenshot 2023-05-31 at 11 23 06 AM

## Type of change

- [x] Bugfix

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge


rohitkrai03 (approved):
/lgtm

PR #683, opened 2023-06-06
Set signup flag to false if any error

karelhala:
## Fixes 

Not able to navigate if signup request fails.

Align the usage of routes, we can gate all routes behind overview if signup returns false.

## Description

If signup URL returns error (401 for instance) the navigation is shown, however none of the routes are properly working as we are not setting the signup to false. This can happen if user tries to access the UI on console.stage.redhat.com environment as this environment is not setup with prod SSO.

Since we can utilize `exact: false` when routing we can hide behind disallowed flag default overview page which'll consume all requests.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

### DIFF
--- a/config/remotePlugin.js
+++ b/config/remotePlugin.js
@@ -127,7 +127,7 @@ const routeExtensions = [
     },
   },
 
-  // application-pipeline overview
+  // application-pipeline overview (use exact if user has signup)
   {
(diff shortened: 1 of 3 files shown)


rohitkrai03 (approved):
/lgtm

PR #696, opened 2023-06-16
Add ephemeral environment dropdown

abhinandan13jan:
## Fixes 

https://issues.redhat.com/browse/HAC-4171

## Description

Adds environment dropdown to integration test creation form

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes
- [ ] Other (please describe)

Screenshot 2023-06-26 at 5 28 29 PM
Screenshot 2023-06-26 at 5 28 50 PM
Screenshot 2023-06-26 at 5 28 57 PM
Screenshot 2023-06-26 at 5 32 53 PM

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/IntegrationTest/IntegrationTestForm/IntegrationTestSection.tsx
+++ b/src/components/IntegrationTest/IntegrationTestForm/IntegrationTestSection.tsx
@@ -8,12 +8,15 @@ import {
 } from '@patternfly/react-core';
 import { useFormikContext, useField } from 'formik';
 import gitUrlParse from 'git-url-parse';
-import { CheckboxField, InputField } from '../../../shared';
+import { useAllEnvironments } from '../../../hooks/useAllEnvironments';
+import { CheckboxField, DropdownField, InputField } from '../../../shared';
 import { useDebounceCallback } from '../../../shared/hooks/useDebounceCallback';
+import { EnvironmentType, getEnvironmentType } from '../../Environment/environment-utils';
 import { AccessHelpText } from '../../ImportForm/SourceSection/SourceSection';
 import { useAccessCheck } from '../../ImportForm/utils/auth-utils';
 import { gitUrlRegex, RESOURCE_NAME_REGEX_MSG } from '../../ImportForm/utils/validation-utils';
-import { IntegrationTestFormValues } from './types';
+import { ENVIRONMENTS, IntegrationTestFormValues } from './types';
+import './IntegrationTestSection.scss';
 
 type Props = { isInPage?: boolean; edit?: boolean };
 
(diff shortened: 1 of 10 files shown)


src/components/IntegrationTest/IntegrationTestForm/IntegrationTestSection.tsx
```diff
@@ -144,10 +176,20 @@ const IntegrationTestSection: React.FC<Props> = ({ isInPage, edit }) => {
...
           data-test="git-path-repo"
           required
         />
+        <DropdownField
+          name={'integrationTest.environmentName'}
+          label="Environment"
+          placeholder={ITSEnvName ?? ENVIRONMENTS.DEFAULT}
+          onChange={setEnvironment}
+          items={dropdownItems}
+          helpText="Select an environment for your test to run on. It can not be changed later"
```
dirgim, line 190: I think that it should be possible to change the base environment after the fact.
The text should also probably need to make clear that the environment that's used for testing will be an ephemeral copy of the selected environment, something like this (but feel free to refine it further):
```
Select an existing environment that the ephemeral test environment for this test  will be based on.
```

abhinandan13jan: I copied the text from the UX copy. cc: @MariaLeonova @beaumorley input here pls

wolfeallison: "Select an existing environment for us to make a copy of. We'll use the copy as a test environment and then delete it." How about this? Tried to make it sound a little more conversational @dirgim @abhinandan13jan

beaumorley: If you can change the environment later, we should update the UX copy. What about something like this?

_Select a static environment for your test to run on. This will be used for the ephemeral test environment._

mattreid:
@abhinandan13jan Can the dropdown for environment selection be the same width as the text fields?

mattreid:
What's the default state for the new environment field when adding a new test? Is it "Development", "No environment", or nothing is selected and the user has to pick something?

src/components/IntegrationTest/IntegrationTestForm/IntegrationTestSection.tsx
mattreid: Thanks for calling that out, Kruno. I had a similar comment in a Miro for these. I think the suggestions here are on the right track. It doesn't all have to go directly into the UI, some detail could be reserved for the docs, but other aspects here I could see being helpful to some folks is clarifying what it means "to copy" (config and env variables will be reused and will match the state of the selected env at the time it is copied), this will be created and deleted each time this test is invoked, an environment doesn't have to be used, they can select no environment (how do they know when they can skip it?).

abhinandan13jan:
@mattreid It is `no environment` by default, fixed the width to match inputField as well
Screenshot 2023-06-21 at 3 56 34 PM

src/components/IntegrationTest/IntegrationTestForm/IntegrationTestSection.tsx
abhinandan13jan: @wolfeallison @beaumorley updated

src/components/IntegrationTest/IntegrationTestForm/IntegrationTestSection.tsx
```diff
@@ -23,13 +25,44 @@ const IntegrationTestSection: React.FC<Props> = ({ isInPage, edit }) => {
     type: 'input',
   });
 
+  const [, { value: ITSEnvName }] = useField<string>({
+    name: 'integrationTest.environmentName',
+    type: 'dropdown',
+  });
+  const [environments, environmentsLoaded] = useAllEnvironments();
```
rohitkrai03, line 32: I am not sure if BYOC environment is supported for ephemeral environment. So you'll need to filter out the static environments for this dropdown.

src/components/IntegrationTest/IntegrationTestForm/IntegrationTestView.tsx
```diff
@@ -44,6 +44,8 @@ const IntegrationTestView: React.FunctionComponent<IntegrationTestViewProps> = (
       url: url?.value ?? '',
       revision: revision?.value ?? '',
       path: path?.value ?? '',
+      environmentName: integrationTest?.spec?.environment?.name,
```
rohitkrai03, line 47: You'll need to set default values for these fields as well.

rohitkrai03 (changes requested):

abhinandan13jan:
@rohitkrai03 updated
1. new description text
2. defaults for environment name and type
3. filtering to use only static and default environments

src/components/IntegrationTest/IntegrationTestForm/IntegrationTestSection.tsx
abhinandan13jan: updated to filter out managed envs

rohitkrai03 (approved):
/lgtm

PR #700, opened 2023-06-20
Add e2e test for default integration test

lcarva:
## Description
Add a test to ensure the default integration test is created automatically when a new Application is created.

https://issues.redhat.com/browse/HACBS-2193

## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Build related changes
- [x] Other (please describe): e2e tests

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/integration-tests/support/pages/tabs/IntegrationTestsTabPage.ts
+++ b/integration-tests/support/pages/tabs/IntegrationTestsTabPage.ts
@@ -16,6 +16,10 @@ export class IntegrationTestsTabPage {
     cy.get(integrationTestsTabPO.filterInputField).clear().type(inputString);
   }
 
+  hasIntegrationTest(integrationTestName: string) {
+    UIhelper.verifyRowInTable('Integration tests', integrationTestName, [integrationTestName]);
+  }
+
   openAndClickKebabMenu(integrationTestName: string, option: string) {
     cy.get(`[data-id="${integrationTestName}"]`).find(actions.kebabButton).click();
     cy.contains('li', option).click();
--- a/integration-tests/tests/basic-happy-path.spec.ts
+++ b/integration-tests/tests/basic-happy-path.spec.ts
@@ -10,10 +10,12 @@ import { Common } from '../utils/Common';
 import { UIhelper } from '../utils/UIhelper';
 import { APIHelper } from '../utils/APIHelper';
 import { FULL_APPLICATION_TITLE } from '../support/constants/PageTitle';
+import { IntegrationTestsTabPage } from '../support/pages/tabs/IntegrationTestsTabPage';
 
 describe('Basic Happy Path', () => {
   const applicationName = Common.generateAppName();
   const applicationDetailPage = new ApplicationDetailPage();
+  const integrationTestsTab = new IntegrationTestsTabPage();
   const publicRepo = 'https://github.com/hac-test/devfile-sample-code-with-quarkus';
   const componentName: string = Common.generateAppName('java-quarkus');
   const piplinerunlogsTasks = ['init', 'clone-repository', 'build-container', 'show-summary'];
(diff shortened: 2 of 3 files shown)


integration-tests/support/pages/tabs/IntegrationTestsTabPage.ts
```diff
@@ -16,6 +16,10 @@ export class IntegrationTestsTabPage {
     cy.get(integrationTestsTabPO.filterInputField).clear().type(inputString);
   }
 
+  hasIntegrationTest(integrationTestName: string) {
+    cy.get(`[data-id="${integrationTestName}"]`);
```
subhashkhileri, line 20: we could move locator to `integrationTestsTabPO`.
Please add `be.visible` and `have.text` assertions.

integration-tests/tests/basic-happy-path.spec.ts
```diff
@@ -41,6 +41,9 @@ describe('Basic Happy Path', () => {
       'Build Running',
       'Default',
     );
+    Applications.goToIntegrationTestsTab().hasIntegrationTest(
```
subhashkhileri, line 44: would prefer new test `it()` for it. can we create new test here for Integration test validation?

integration-tests/utils/Applications.ts
```diff
@@ -142,6 +143,7 @@ export class Applications {
 
   static goToIntegrationTestsTab() {
     cy.get(integrationTestsTabPO.clickTab).click();
+    return new IntegrationTestsTabPage();
```
subhashkhileri, line 146: 🤔 Instead of returning instance here can we import it in basic flow? may be for any other validations we can utilise it if there is a requirement.

lcarva: Sounds good! Updated. (It looks like half of the `goTo*` functions return a `*Page` instance, while the other half does not. I picked the wrong pattern to follow :smile: )

integration-tests/support/pages/tabs/IntegrationTestsTabPage.ts
lcarva: Done!

lcarva:
@subhashkhileri I believe I have address your comments :pray:

integration-tests/support/pages/tabs/IntegrationTestsTabPage.ts
```diff
@@ -16,6 +16,13 @@ export class IntegrationTestsTabPage {
     cy.get(integrationTestsTabPO.filterInputField).clear().type(inputString);
   }
 
+  hasIntegrationTest(integrationTestName: string) {
+    const tr = cy.get(integrationTestsTabPO.item(integrationTestName)).should('be.visible');
+    tr.get(integrationTestsTabPO.itemName)
```
subhashkhileri, line 21: cypress commands run asynchronously, so assigning it to variable is probably not a right way you can update to : 
`cy.get(integrationTestsTabPO.item(integrationTestName)).should('be.visible').should('have.text', integrationTestName)`

Hopefully this should work.

integration-tests/utils/Applications.ts
```diff
@@ -21,6 +21,7 @@ import { FULL_APPLICATION_TITLE } from '../support/constants/PageTitle';
 import { UIhelper } from './UIhelper';
 import { hacAPIEndpoints } from '../utils/APIEndpoints';
 import { APIHelper } from '../utils/APIHelper';
+import { IntegrationTestsTabPage } from '../support/pages/tabs/IntegrationTestsTabPage';
```
subhashkhileri, line 24: we can remove `IntegrationTestsTabPage` import if its not being used.

integration-tests/support/pages/tabs/IntegrationTestsTabPage.ts
lcarva: I'm having trouble running the tests now. It halts when creating the application. I'll try this approach when I sort that out.

lcarva: Worked like a charm! Thanks for the suggestion. Updated.

lcarva:
@subhashkhileri ready for another look!

PR #718, opened 2023-07-06
Add support for Tekton result logs

jeff-phillips-18:
## Fixes 
Fixes https://issues.redhat.com/browse/HAC-4318

## Description
Adds support for retrieving log results from Texton when pods are no longer available.

## Type of change
- [x] Feature

image

/cc @rohitkrai03 @christianvogt


christianvogt:
works as expected.

Downloading all logs takes a while. I wonder if we ought to have a tooltip popup and show additional information. However this, if at all, should be part of a separate PR.

jeff-phillips-18:
@christianvogt @rohitkrai03 I'm not able to increase the code coverage. The non-covered code is refactored to be asynchronous vs synchronous and it's dependent on reading logs and waiting for them to be returned.

PR #729, opened 2023-07-21
Fetch pipelineruns only for active components in an application

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/RHTAPBUGS-483

## Description

Tekton results query for fetching pipelineruns returns all the records that matches the application, this leads to showing the pipelineruns for the components which was deleted long before.

**Solution:**

 Updated the query to fetch the pipelinruns by passing the active components  in the application to avoid showing old invalid build pipelineruns.
 

## Type of change

- [x] Bugfix

**Before:** 

Shows deleted Component's build pipelinerun in the list view:
image

**After:**

Shows only the build pipeline runs of all active components.

image

## How to test or reproduce?

The pipelinerun list view should not include the deleted component's pipelinerun

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

cc: @rohitkrai03


rottencandy (approved):
/lgtm

PR #736, opened 2023-07-27
add workspace context in the tekton results api requests

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/RHTAPBUGS-522

## Description

Tekton results API is failing with `401` Unauthorised error while you are in a shared workspace and this causes UI to not load  visualisations and component list view pages. This happens because the workspace information is not included in the API call so it defaults to user's home workspace.

Solution:
This PR appends workspace information in the tekton-results API request to properly map to a correct workspace.

All tekton-results API calls should contain workspace context eg :
`workspaces/<worksspace-name>/apis/results.tekton.dev/v1alpha2/parents/<namespace>/results/-/records`

## Type of change

- [x] Bugfix

## How to test or reproduce?

1. Switch to a shared workspace and visit application details page or components list pages.

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

### DIFF
--- a/src/components/LogViewer/BuildLogViewer.tsx
+++ b/src/components/LogViewer/BuildLogViewer.tsx
@@ -88,7 +88,7 @@ export const BuildLogViewer: React.FC<BuildLogViewerProps> = ({ component }) => 
       </div>
       <div className="build-log-viewer__body">
         {pipelineRun && taskRuns && tloaded ? (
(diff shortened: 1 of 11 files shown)


rottencandy (approved):
/lgtm

PR #741, opened 2023-07-31
Add check for all buttons to be ready.

Katka92:
## Fixes 
https://issues.redhat.com/browse/HAC-4543

## Description
Tests are stuck on a loading page when the `Applications` are clicked after the login. Add a check for the `Get started` page to be fully loaded to prevent it.

### DIFF
--- a/integration-tests/utils/Login.ts
+++ b/integration-tests/utils/Login.ts
@@ -1,5 +1,6 @@
 import { NavItem, pageTitles } from '../support/constants/PageTitle';
 import { loginPO, kcLoginPO } from '../support/pageObjects/global-po';
+import { GetStartedPage } from '../support/pages/GetStartedPage';
 import { Common } from './Common';
 
 export class Login {
@@ -29,7 +30,7 @@ export class Login {
 
   private static waitForApps() {
     Common.waitForLoad();
-    Common.verifyPageTitle(pageTitles.overviewPage);
+    GetStartedPage.waitForLoad();
     Common.navigateTo(NavItem.applications);
     Common.verifyPageTitle(pageTitles.applications);
     Common.waitForLoad();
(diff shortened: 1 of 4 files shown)


integration-tests/utils/Login.ts
```diff
@@ -30,6 +30,10 @@ export class Login {
   private static waitForApps() {
     Common.waitForLoad();
     Common.verifyPageTitle(pageTitles.overviewPage);
+    cy.get('[data-test="view-my-applications"]')
+      .should('be.visible')
+      .and('have.attr', 'aria-disabled')
```
subhashkhileri, line 35: I guess `.should('be.enabled')` should also work.

Katka92: Do you mean instead of checking `aria-disabled=false` I should just use this `be.enabled` check?

Katka92: I tried locally: 
```
    cy.get('[data-test="view-my-applications"]')
      .should('be.visible')
      .and('be.enabled')
```
and also
```
    cy.get('[data-test="view-my-applications"]')
      .should('be.enabled')
```
but failed with: 
```
Timed out retrying after 40000ms: expected '<a.pf-c-button.pf-m-secondary.pf-m-display-lg.intro-banner__cta>' to be 'enabled'
```
Not sure why it is checking this button instead of the element I get by `cy.get()`.

subhashkhileri: its checking  <a> tag only not button as per the above failure.

integration-tests/utils/Login.ts
```diff
@@ -30,6 +30,10 @@ export class Login {
...
+      .and('equal', 'false');
```
subhashkhileri, line 36: if be.enabled isn't working then I am ok with this.
can we move 'false' to  `.and('have.attr', 'aria-disabled', 'false')` it should also work.

PR #745, opened 2023-08-02
Improve performance by batching tekton-results API requests in PLR list pages.

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/RHTAPBUGS-545

## Description

PipelineRun list view was making too many calls to tekton-results API and the users were hitting rate limit issue in the prod. 

This PR improves the performance of the list page with following changes

1. Batch request the vulnerability calls (added the fetching logic to get vulnerability data for 30 rows in a table in a single API call)
2. reduce the pageSize from `50` -> `30`
3. do not fetch extra calls if the same request is already inflight.

With this changes, each pipelineruns list row will not make any api calls instead the parent will make a single call to fetch all the vulnerabilities and set them accordingly.

## Type of change

- [x] Other (please describe): No visual changes, this PR improves the performance of pipelinerun list page and avoid api rate limit issue.

Before:

https://github.com/openshift/hac-dev/assets/9964343/4bb2e708-66e5-4ddc-bc00-be2070909480

After:

https://github.com/openshift/hac-dev/assets/9964343/3d38306e-7fbc-439c-adc6-73e926117d88

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

### DIFF
--- a/src/components/Commits/tabs/CommitsPipelineRunTab.tsx
+++ b/src/components/Commits/tabs/CommitsPipelineRunTab.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { Title } from '@patternfly/react-core';
+import { usePLRVulnerabilities } from '../../../hooks/useScanResults';
 import { Table } from '../../../shared';
 import { PipelineRunKind } from '../../../types';
 import PipelineRunEmptyState from '../../PipelineRunDetailsView/PipelineRunEmptyState';
@@ -15,6 +16,8 @@ const CommitsPipelineRunTab: React.FC<CommitsPipelineRunTabProps> = ({
   pipelineRuns,
   applicationName,
 }) => {
+  const vulnerabilities = usePLRVulnerabilities(pipelineRuns);
+
   if (!pipelineRuns || pipelineRuns.length === 0) {
     return <PipelineRunEmptyState applicationName={applicationName} />;
   }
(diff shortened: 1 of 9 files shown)


rohitkrai03:
You can see that just within a few seconds of scrolling up and down it fired almost 900 API requests. 
Screenshot 2023-08-03 at 9 32 23 PM

karthikjeeyar:
@rohitkrai03 By looking at the video I can confirm that this is a old behaviour, and looks like it doesn't contain my PR changes. If you notice while you scroll up and down fastly, the vulnerabilities column indicates it is loading but with my PR changes once the column is rendered then even if you scroll up or down on that row again and again it will never go back to the loading state. I couldn't reproduce this with my changes, could you please retry again?

karthikjeeyar:
Pushed minor changes to show loading indicator for all inflight API calls and added some more unit tests.

rohitkrai03:
> @rohitkrai03 By looking at the video I can confirm that this is a old behaviour, and looks like it doesn't contain my PR changes. If you notice while you scroll up and down fastly, the vulnerabilities column indicates it is loading but with my PR changes once the column is rendered then even if you scroll up or down on that row again and again it will never go back to the loading state. I couldn't reproduce this with my changes, could you please retry again?

@karthikjeeyar Yup, seems like my local was not running the latest build from your PR. After refreshing it seems to be working much better and the issue of infinite calls during scrolling is fixed completely.

PR #758, opened 2023-08-08
reset active workspace before listing workspaces

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/RHTAPBUGS-568
https://issues.redhat.com/browse/RHTAPBUGS-564

## Description

Fetching workspaces returns empty array due to the invalid url.

RCA:

UI makes calls to list the workspaces and sets the active namespace and workspace, but for this call to work properly we should not set the workspace context prior to this call or else this will result in empty array and impact all other subsequent API calls that UI makes.

Solution:

Reset the active workspace before list workspaces

## Type of change

- [x] Bugfix

**Before:**

empty_respsone incorrect-workspace-url

**After**:

https://github.com/openshift/hac-dev/assets/9964343/84dd0401-3d2e-4a5b-b1f5-3319af2ce784

Screenshot 2023-08-08 at 12 20 05 PM

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

### DIFF
--- a/src/components/PipelineRunListView/PipelineRunListRow.tsx
+++ b/src/components/PipelineRunListView/PipelineRunListRow.tsx
@@ -53,8 +53,17 @@ const BasePipelineRunListRow: React.FC<BasePipelineRunListRowProps> = ({
         />
       </TableData>
       {showVulnerabilities ? (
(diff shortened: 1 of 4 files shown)


rohitkrai03 (approved):
/lgtm

PR #761, opened 2023-08-09
update TR Api query filter and page size in lifecycle graph 

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/RHTAPBUGS-574

## Description

The default limit for a response in tekton results is 4mb but the lifecycle graph makes call with `100` pagesize so the response is more than the limit and fails to load the graph.

Included the component names in the query filter and also dynamically setting the pageSize based on the number of components in an application.

## Type of change

- [x] Bugfix

**Before**:

Screenshot 2023-08-09 at 4 00 17 PM
Screenshot 2023-08-09 at 4 00 27 PM

**After**:

After_fix

## How to test or reproduce?

Visit acm shared workspace application details page to see the graph loading.

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge


rohitkrai03 (approved):
/lgtm

PR #783, opened 2023-08-30
Fix for recent commits rows disappearing on scroll

jeff-phillips-18:
## Fixes 
Fixes [RHTAPBUGS-730](https://issues.redhat.com/browse/RHTAPBUGS-730)

## Description
Commits on the application overview page disappear on scroll due to the virtualization computing the table location which can change once the data is loaded (no longer showing the spinner) or if the user changes the graph from collapsed to expanded.

There is no need to virtualize this table since there will be at most 5 rows shown.

Update the `Table` component to correctly handle tables that choose not to be `virtualized` by adding the rows directly rather than using the WindowScroller and VirtualBody components.

## Type of change
- [x] Bugfix


karthikjeeyar:
@jeff-phillips-18 It looks like e2e test needs to be updated to use `table` instead of `div`with this change, but I wonder for other virtualized tables like `Pipeline runs` list it will be still `div` 🤔 

Screenshot 2023-08-31 at 1 40 02 PM

PR #794, opened 2023-09-06
fix: consider all pipeline runs, by using paging

jeff-phillips-18:
## Fixes 
Fixes [RHTAPBUGS-752](https://issues.redhat.com/browse/RHTAPBUGS-752)

## Description
When retrieving pipeline runs, call getNextPage if it is returned in order to keep getting results until we have them all.

## Type of change
- [x] Bugfix


src/components/Commits/CommitsListView.tsx
```diff
@@ -77,6 +77,12 @@ const CommitsListView: React.FC<CommitsListViewProps> = ({
     ),
   );
 
+  React.useEffect(() => {
+    if (loaded && !plrError && getNextPage) {
+      getNextPage();
+    }
+  }, [loaded, plrError, getNextPage]);
```
karthikjeeyar, line 84: This simply over fetches all the pipelineruns in the workspace even if the user isn't scrolling down the commit list view. I think its better to fire this `getNextPage` only when the User hits the bottom of the table just like we do for pipelinerunListView.

jeff-phillips-18: The fetching on page scroll would work only if the commits were paged. If we don't get all the pipeline runs, we may not find the run associated with a commit that is shown.

jeff-phillips-18:
@karthikjeeyar This should be good to go. PTAL when you have a moment.

karthikjeeyar:
For a workspace that has larger data set the overview visualization takes too much time to resolve. I have a workspace where it takes approx 2minutes to resolve. not sure if we can do anything to improve, we could load the visualization as soon as we receive the first set of data and update it as we get more, but it wont be accurate. wdyt?

jeff-phillips-18:
> For a workspace that has larger data set the overview visualization takes too much time to resolve. I have a workspace where it takes approx 2minutes to resolve. not sure if we can do anything to improve, we could load the visualization as soon as we receive the first set of data and update it as we get more, but it wont be accurate. wdyt?

I've made a couple of updates so that we only fetch data if all the records do not have a match and to fetch only data for the rows that we don't yet have a match for. So the set of components used to filter by is reduced to only the components we do not yet have the latest build pipeline run for. This prevents a lot of fetching. I have it down from 2 minutes to about 15 seconds.

PR #808, opened 2023-09-21
Add CDQ pact test

Katka92:
## Fixes 
A test we added during a Workshop.

## Description
Testing the CDQ creation response.


pact-tests/contracts/application-service/cdq-creation.ts
```diff
@@ -0,0 +1,77 @@
...
+  metadata: {
+    creationTimestamp: regex(matchers.dateAndTime, '2022-01-21T13:36:30Z'),
+    generation: like(1),
+    name: cdq,
+    namespace,
+  },
+  spec: {
+    git: {
+      url: 'https://github.com/nodeshift-starters/devfile-sample.git',
+    },
```
karthikjeeyar, line 47:
```suggestion
    git: {
      context: '',
      revision: '',
      url: 'https://github.com/nodeshift-starters/devfile-sample.git',
    },
```

pact-tests/contracts/application-service/cdq-creation.ts
```diff
@@ -0,0 +1,77 @@
...
+    conditions: [
+      {
+        lastTransitionTime: regex(matchers.dateAndTime, '2022-06-01T19:46:56Z'),
+        message: like('ComponentDetectionQuery is processing'),
+        reason: 'Success',
+        status: 'True',
+        type: 'Processing',
+      },
+    ],
+  },
```
karthikjeeyar, line 59: I have checked this creation scenario and the `status` field wasn't added immediately after the POST call. After reconciliation it adds this status field so maybe we can move this to a new `get-cdq` contract

karthikjeeyar (commented):
After the Pact workshop, I had verified the cdq creation and get call. I have added some comments below based on my observation.

pact-tests/contracts/application-service/cdq-creation.ts
Katka92: So, if I get that, you want to check that the fields `context` and `revision` are there and you also want to check that they are empty?

PR #812, opened 2023-09-26
Update computation of deployed application status

jeff-phillips-18:
## Fixes 
Fixes [RHTAPBUGS-783](https://issues.redhat.com/browse/RHTAPBUGS-783)

## Description
Update the computation of the deployed application status to use status.gitopsDeployment instead of status.componentsDeploymentConditions

## Type of change
- [x] Bugfix


rohitkrai03 (approved):
/approve

jeff-phillips-18:
Updated application header to match status show in the cards (though this may be removed soon at least it is consistent until it is removed):

image

PR #826, opened 2023-10-02
fix(task logs): show logs for skipped or idle tasks if they exist

jeff-phillips-18:
## Fixes 
Fixes [HAC-5103](https://issues.redhat.com/browse/HAC-5103)

## Description
Attempt to get pod logs for skipped or idle tasks. Show existing messages only if the pod does not exist.

## Type of change
- [x] Bugfix


rottencandy (approved):
/lgtm

PR #841, opened 2023-10-25
fix(secret-list): fix secret list page error when secret is missing status field

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/RHTAPBUGS-913

## Description

Secret list page is throwing error when it is redirected from Add secret page. recently created secret is not having status field initially and the list page is looking for the status field to read the number of key/value pairs in the secret which is causing the page to fail.

## Type of change

- [x] Bugfix

https://github.com/openshift/hac-dev/assets/9964343/8f53643a-d88e-4162-aebd-0376d5158150

## How to test or reproduce?

Add a build secret from the UI and wait for the UI to redirect to the secret list page.

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

### DIFF
--- a/src/components/Secrets/__tests___/secret-utils.spec.ts
+++ b/src/components/Secrets/__tests___/secret-utils.spec.ts
@@ -160,6 +160,24 @@ describe('getSecretRowData', () => {
     });
   });
 
+  it('should not throw error when the status field is missing in the newly created secret', () => {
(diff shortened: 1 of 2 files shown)


rottencandy (approved):
/lgtm

PR #847, opened 2023-10-30
fix(snapshot): add test pipelinerun list and capitalize breadcrumbs

abhinandan13jan:
## Fixes 

https://issues.redhat.com/browse/RHTAPBUGS-839

## Description

- Updates PLR list to include test PLRs with snapshot label
- capitalizes snapshot in breadcrumb
- capitalizes Environmenr names

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

Screenshot 2023-10-31 at 6 26 16 PM
Screenshot 2023-10-31 at 6 26 25 PM

Screenshot 2023-10-30 at 8 16 24 PM
Screenshot 2023-10-30 at 8 16 39 PM

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/SnapshotDetails/SnapshotDetailsView.tsx
+++ b/src/components/SnapshotDetails/SnapshotDetailsView.tsx
@@ -112,7 +112,7 @@ const SnapshotDetailsView: React.FC<React.PropsWithChildren<SnapshotDetailsViewP
           ...applicationBreadcrumbs,
           {
             path: `#`,
-            name: 'snapshots',
+            name: 'Snapshots',
           },
           {
             path: `/application-pipeline/workspaces/${workspace}/applications/${applicationName}/snapshots/${snapshotName}`,
@@ -146,7 +146,7 @@ const SnapshotDetailsView: React.FC<React.PropsWithChildren<SnapshotDetailsViewP
                       <>
                         <StatusIconWithTextLabel
                           key={env.metadata?.name}
-                          text={env.metadata.name}
+                          text={env.spec.displayName ?? env.metadata.name}
                           dataTestAttribute="snapshot-env-label"
                           status={isDeployed ? runStatus.Succeeded : runStatus.Cancelling}
                         />{' '}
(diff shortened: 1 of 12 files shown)


mattreid:
Is Type being lowercase in the filter selector coming from data, or our code? Other areas we're filtering on PLR type, it shows as capitalized (Build, Test).

abhinandan13jan:
@mattreid we get it as lowercase types from the object metadata. I updated to capitalize it while displaying
Screenshot 2023-10-31 at 6 26 25 PM

src/components/SnapshotDetails/SnapshotDetailsView.tsx
```diff
@@ -147,7 +147,7 @@ const SnapshotDetailsView: React.FC<SnapshotDetailsViewProps> = ({
                       <>
                         <StatusIconWithTextLabel
                           key={env.metadata?.name}
-                          text={env.metadata.name}
+                          text={capitalize(env.metadata.name)}
```
karthikjeeyar, line 150: There is a `env.spec?.displayName` optional field in Environment CR, We must use that if exists or else fallback to the env.metadata.name. This way we will be be respecting the user's choice of displayName rather than Capitalizing the K8s resource name.

karthikjeeyar: Deployment tab uses the above approach, so the `Development` text comes from spec.displayName and notice the other environment which doesn't have displayName and it uses resource name. Can we stick to this displayName or resource.name (without captializing) approach for consistency ? cc: @mattreid 


image

src/components/SnapshotDetails/tabs/SnapshotOverview.tsx
```diff
@@ -108,7 +109,7 @@ const SnapshotOverviewTab: React.FC<SnapshotOverviewTabProps> = ({
                           <Link
                             to={`/application-pipeline/workspaces/${workspace}/applications/${snapshot.spec.application}/deployments`}
                           >
-                            {env}
+                            {capitalize(env)}
```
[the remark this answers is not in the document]
abhinandan13jan: updated

rottencandy (approved):
Looking good, but looks like we need more tests to satisfy codecov

PR #852, opened 2023-11-06
fix(e2e): select the correct itest pipeline

jrichter1:
## Fixes 
https://issues.redhat.com/browse/RHTAPBUGS-951

## Description
Make sure the tests end up on the push triggered pipeline run when checking itests


jrichter1:
[test]

PR #888, opened 2024-01-03
fix(import-flow): close submit progress model before customize pipeline

sahil143:
## Fixes 

https://issues.redhat.com/browse/RHTAPBUGS-1084

## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Build related changes

Kapture 2024-01-03 at 16 31 42

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/ImportForm/GitImportForm.tsx
+++ b/src/components/ImportForm/GitImportForm.tsx
@@ -92,6 +92,11 @@ const GitImportForm: React.FunctionComponent<React.PropsWithChildren<GitImportFo
 
   const showModal = useModalLauncher();
 
+  const closeSubmitProgressModal = () => {
+    setSubmitModalText('');
+    setValueProgressBar(0);
+  };
+
   const handleSubmit = React.useCallback(
     (values: ImportFormValues, formikHelpers: any) => {
       track(TrackEvents.ButtonClicked, { link_name: 'import-submit', workspace });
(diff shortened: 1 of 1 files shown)


rottencandy (approved):
/lgtm

src/components/ImportForm/GitImportForm.tsx
```diff
@@ -110,7 +115,7 @@ const GitImportForm: React.FunctionComponent<React.PropsWithChildren<GitImportFo
         // no. of secrets with dry run requests
         (values?.importSecrets.length ?? 0) * 2;
       setProgressBarData({ maxValueForProgressBar, noOfComponents });
-      return createResources(values, ImportStrategy.GIT, workspace)
+      createResources(values, ImportStrategy.GIT, workspace)
```
rohitkrai03, line 118: If you don't return the promise to formik it will not set proper state values like isSubmitting.

rohitkrai03 (changes requested):

src/components/ImportForm/GitImportForm.tsx
rohitkrai03: @sahil143 ^^

sahil143: We manually set the submitting state in case of an error and if there is no error it will redirect. There is a delay in opening the custom pipeline modal during which the create application button is enabled because promise is resolved and formik is setting isSumbitting to true. To prevent this behaviour I removed the return statement.

rohitkrai03 (approved):
/lgtm

PR #891, opened 2024-01-04
fix(overview): set max height for tooltips on nodes

jeff-phillips-18:
## Fixes 
Fixes [RHTAPBUGS-1080](https://issues.redhat.com/browse/RHTAPBUGS-1080)

## Description
The height of the tooltips was not constrained at all so if there are a lot of builds (or any other items) in a tooltip it would be larger than the view height causing the window to scroll past the current view leaving lots of blank space.

Set the maximum height of the tooltip to be 600 pixels and have the individual items in a scrollable area so the user can view all the items if desired.

## Type of change
- [x] Bugfix

image

## How to test or reproduce?
Create a project with many many builds. Hover over the build node in the overview section.


rohitkrai03 (approved):
/lgtm

PR #896, opened 2024-01-12
fix(logs): Update PLR logs fetch logic to match the new tekton results changes

karthikjeeyar:
## Fixes 

https://issues.redhat.com/browse/RHTAPBUGS-1093

## Description

PipelineRun logs is not streaming properly in the UI because the tekton results logs API is now returning raw logs content directly, so UI doesn't need to decode the value again.

tekton-results PR introduced this change - https://github.com/tektoncd/results/pull/632

## Type of change

- [x] Bugfix

Before:

image

After:

image

## How to test or reproduce?

- Visit logs tab in the pipelinerun details page

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

### DIFF
--- a/src/utils/tekton-results.ts
+++ b/src/utils/tekton-results.ts
@@ -1,5 +1,6 @@
 import {
   commonFetchJSON,
+  commonFetchText,
   K8sResourceCommon,
   MatchExpression,
   MatchLabels,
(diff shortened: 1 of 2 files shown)


src/utils/tekton-results.ts
```diff
@@ -329,10 +330,11 @@ export const getTaskRuns = (
   cacheKey?: string,
 ) => getFilteredTaskRuns(workspace, namespace, '', options, nextPageToken, cacheKey);
 
-const getLog = (workspace: string, taskRunPath: string) =>
-  commonFetchJSON<Log>(
+const getLog = async (workspace: string, taskRunPath: string) => {
```
jeff-phillips-18, line 333: nit: No reason to make this async

karthikjeeyar: updated

PR #898, opened 2024-01-30
feat(Component-relationship): add modal to define component relationship

sahil143:
## Fixes 

https://issues.redhat.com/browse/HAC-5432

## Description

add component relation modal for applications

TODO:
- [ ] add unit tests 

## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Build related changes

Kapture 2024-01-30 at 11 14 22

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/ComponentRelation/ComponentRelationshipDropdowns.tsx
+++ b/src/components/ComponentRelation/ComponentRelationshipDropdowns.tsx
@@ -1,0 +1,190 @@
+import * as React from 'react';
+import {
+  Badge,
+  Divider,
+  Menu,
+  MenuContainer,
+  MenuContent,
+  MenuItem,
+  MenuList,
+  MenuToggle,
+} from '@patternfly/react-core';
+import { useField } from 'formik';
+
+type SelectComponentsDropdownProps = {
+  children: React.ReactNode[];
+  toggleText: string;
+  onSelect: (item: string | number) => void;
+  closeOnSelect?: boolean;
+  badgeValue?: number;
+};
+
+const SelectComponentsDropdown: React.FC<SelectComponentsDropdownProps> = ({
+  children,
+  toggleText,
+  onSelect,
+  closeOnSelect,
+  badgeValue,
+}) => {
+  const [isOpen, setIsOpen] = React.useState(false);
+  const toggleRef = React.useRef<HTMLButtonElement>(null);
+  const menuRef = React.useRef<HTMLDivElement>(null);
+  return (
+    <MenuContainer
+      isOpen={isOpen}
+      onOpenChange={(o) => setIsOpen(o)}
+      toggle={
+        <MenuToggle
+          ref={toggleRef}
+          isExpanded={isOpen}
+          onClick={() => setIsOpen(!isOpen)}
+          id="toggle-component-menu"
+          aria-label="toggle component menu"
+          badge={badgeValue ? <Badge isRead>{badgeValue}</Badge> : null}
+          isFullWidth
+        >
+          {toggleText}
+        </MenuToggle>
+      }
+      menu={
+        <Menu
+          ref={menuRef}
+          id="multi-select-component-menu"
+          isScrollable
+          onSelect={(_, item) => {
+            onSelect(item);
+            closeOnSelect && setIsOpen(false);
+          }}
+        >
+          <MenuContent>
+            <MenuList>{children}</MenuList>
+          </MenuContent>
+        </Menu>
+      }
+      toggleRef={toggleRef}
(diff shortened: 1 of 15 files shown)


MariaLeonova:
@karthikjeeyar yes, I think if there's only one, there's no need for the Define relationships option to be active

src/components/ComponentRelation/useNudgeData.ts
```diff
@@ -0,0 +1,35 @@
...
+            return [...acc, data];
+          }
+          if (val.status?.[NudgeStats.NUDGED_BY]) {
+            const data: ComponentRelationValue = {
+              source: val.metadata.name,
+              nudgeType: ComponentRelationNudgeType.NUDGED_BY,
+              target: val.status?.[NudgeStats.NUDGED_BY],
+            };
+            return [...acc, data];
+          }
```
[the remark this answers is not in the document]
arewm: I agree with @MariaLeonova . If possible, we should only show the relationships that users set and not the auto-generated inverse ones.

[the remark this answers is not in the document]
sahil143: @MariaLeonova  If user creates `is nudged by` relationship then also we make a nudges API call behind the scenes. So if user creates a `is nudged by` relationship and then comes back to this modal after saving they will only see `nudges` relationships in the modal.

src/components/ComponentRelation/ComponentRelationshipDropdowns.tsx
```diff
@@ -0,0 +1,190 @@
...
+};
+
+const SelectComponentsDropdown: React.FC<SelectComponentsDropdownProps> = ({
+  children,
+  toggleText,
+  onSelect,
+  closeOnSelect,
+  badgeValue,
+}) => {
+  const [isOpen, setIsOpen] = React.useState(false);
```
abhinandan13jan, line 29: type boolean

src/components/ComponentRelation/cr-modals.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+const SuccessIcon: React.FC = () => (
+  <Icon status="success">
+    <CheckCircleIcon />
+  </Icon>
+);
+
+type ConfirmSubmissionComponentRelationModalProps = Pick<RawComponentProps, 'modalProps'>;
+
+export const ConfirmSubmissionComponentRelationModal: React.FC<
```
abhinandan13jan, line 127: can we move this out to separate file

sahil143:
> There is no going back if I accidentally clicked on the add another relationship button, there is no remove button so I had to close the current process and start it again.

@MariaLeonova can you please take a look at this comment?

src/components/ComponentRelation/useNudgeData.ts
```diff
@@ -0,0 +1,27 @@
...
+  const { namespace } = useWorkspaceInfo();
+  const [components, loaded, error] = useComponents(namespace, application);
+  const nudgeData: ComponentRelationValue[] = React.useMemo(() => {
+    return loaded && !error
+      ? components.reduce((acc, val: ComponentKind) => {
+          if (val.spec?.[NudgeStats.NUDGES]) {
+            const data: ComponentRelationValue = {
+              source: val.metadata.name,
+              nudgeType: ComponentRelationNudgeType.NUDGES,
+              target: val.spec[NudgeStats.NUDGES],
```
karthikjeeyar, line 17: we can take into account that only available components in the application will be used to show relationships.

```suggestion
              target: val.spec[NudgeStats.NUDGES]?.filter(
                (cname) => !!components?.find((co) => co?.metadata?.name === cname),,
```

arewm: This restriction to limit nudges within an Application have actually been loosened. We should no longer filter to only Components in the one Application. 

@MariaLeonova , this has a potential adverse effect on the UX as the number of components can be much larger. 

Would it be reasonable/possible to be able to filter components based on their Application? Or to have Components listed with their Application name (i.e. `<Application>/<Component>`) for grouping "similar" Components together?

sahil143: @arewm We will add support for component relations across applications in a new story.

rohitkrai03 (approved):
/approve

src/components/ComponentRelation/useNudgeData.ts
arewm: @sahil143 , do you have the feature to for cross-Application support?

PR #939, opened 2024-06-03
fix(import): do not send generate annotation in component

rohitkrai03:
## Description

- Do not send generate image annotation since we are creating image repository CR now.

## Type of change

- [x] Bugfix

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


sahil143 (approved):
/lgtm

PR #955, opened 2024-06-26
fix(ImageRepository): add notification data and generate sbom url

sahil143:
## Description

https://issues.redhat.com/browse/HAC-5720
https://issues.redhat.com/browse/HAC-5782

## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Build related changes

image

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


gbenhaim:
@sahil143 does the notification configuration is hardcoded? 
If so, this is not something we can have for the community version.

src/utils/create-utils.ts
```diff
@@ -399,6 +400,16 @@ export const createImageRepository = (
           ? ImageRepositoryVisibility.private
           : ImageRepositoryVisibility.public,
       },
+      notifications: [
```
gbenhaim, line 403: @sahil143 does the notification configuration is hardcoded?
If so, this is not something we can have for the community version.

gbenhaim (changes requested):

src/hooks/useUIInstance.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export const useUIInstance = (): ConsoleDotEnvironments => {
+  const env = useChrome((chrome) => chrome.getEnvironment());
+  return getInternalInstance() ?? (env as ConsoleDotEnvironments);
+};
+
+const SBOM_PLACEHOLDER = '<PLACEHOLDER>';
+const getSBOMEnvUrl = (env: ConsoleDotEnvironments) => (imageHash: string) => {
+  if (env === ConsoleDotEnvironments.prod) {
+    return `https://atlas.devshift.net/sbom/content/${SBOM_PLACEHOLDER}`.replace(
```
gbenhaim, line 33: All of those urls should be discovered from the backend.

gbenhaim (changes requested):

src/components/PipelineRunDetailsView/tabs/PipelineRunDetailsTab.tsx
```diff
@@ -203,17 +208,11 @@ const PipelineRunDetailsTab: React.FC<React.PropsWithChildren<PipelineRunDetails
...
                     <DescriptionListTerm>SBOM</DescriptionListTerm>
                     <DescriptionListDescription>
-                      <Link
-                        to={`/application-pipeline/workspaces/${workspace}/applications/${
-                          sbomTaskRun.metadata.labels[PipelineRunLabel.APPLICATION]
-                        }/taskruns/${sbomTaskRun.metadata.name}/logs`}
-                      >
-                        View SBOM
-                      </Link>
+                      <ExternalLink href={generateSbomUrl(imageDigest)}>View SBOM</ExternalLink>
```
gbenhaim, line 215: This should be conditionally rendered. 
If the Konflux administrator didn't configure a Bambino instance, we shouldn't show a link here.

gbenhaim (changes requested):

Allda:
I see you are using `bambino` in multiple places. The name of the service is `Bombino` and you should fix the typo unless it was made on purpose.

src/components/PipelineRunDetailsView/tabs/PipelineRunDetailsTab.tsx
sahil143: Yes, I agree, but how do we identify whether or not the bombino URL is available?

sahil143: Right now it's just a static URL and we append the image digest at the end of the url.

rohitkrai03: Can we identify the community instance?

sahil143: yes, we can. do we want to send these notifications and show the SBOM URL only for internal instances?

gbenhaim: For now I would say yes. going forward I think that the integration of Konflux with an SBOM server will be valuable also for the community version. Thus, I think that the details of SBOM server should be read from a configmap on the cluster, but this can be done in a different PR in order to unblock this one.

rohitkrai03 (approved):
/approve

src/components/PipelineRunDetailsView/tabs/PipelineRunDetailsTab.tsx
sahil143: There isn't a proper way to identify the internal instance right now. We will figure this out in separate PR

PR #957, opened 2024-06-27
Use new git location for ec pipeline definition

simonbaird:
## Fixes 

* https://issues.redhat.com/browse/EC-668

## Description

The build-definitions repo moved over to the new konflux-ci org in GitHub. Update the pipeline definition git resolver url accordingly.

## Type of change

One-liner to update a git url.

Should be zero visible change.

## How to test or reproduce?

I expect the risk is almost zero here. CI might be enough, but to test it, create a component and make sure the automatically created Enterprise Contract Integration Test Scenario is create and runs as expected.

### DIFF
--- a/src/components/IntegrationTest/IntegrationTestForm/const.ts
+++ b/src/components/IntegrationTest/IntegrationTestForm/const.ts
@@ -1,4 +1,4 @@
-export const EC_INTEGRATION_TEST_URL = 'https://github.com/redhat-appstudio/build-definitions';
+export const EC_INTEGRATION_TEST_URL = 'https://github.com/konflux-ci/build-definitions';
 export const EC_INTEGRATION_TEST_REVISION = 'main';
 export const EC_INTEGRATION_TEST_PATH = 'pipelines/enterprise-contract.yaml';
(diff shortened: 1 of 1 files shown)


sahil143 (approved):

rohitkrai03 (approved):
/lgtm

simonbaird:
This should be an easy-merge I think. Does it really need two more reviews? Anyone want to hit the merge button?

simonbaird:
To confirming the url change, try clicking https://github.com/redhat-appstudio/build-definitions and notice where you end up.

simonbaird:
https://github.com/konflux-ci/build-definitions/blob/main/pipelines/enterprise-contract.yaml is the pipeline being referenced in that repo for the integration test.

zregvart:
The problem here is that the "Red Hat Konflux" check suite is [demanding](https://github.com/openshift/hac-dev/pull/957/checks?check_run_id=26781806178) something from someone, and apparently no one knows what that is:

```
User simonbaird is not allowed to trigger CI via pull_request on this repo.
```

**And** tide is (mis)configured to merge only if "Red Hat Konflux" check suite is marked as successful.

simonbaird:
Thanks @rohitkrai03 .

PR #958, opened 2024-06-28
fix(test): update Pipeline label value in pipeline run details view

Katka92:
## Description
Advanced tests started to fail when verifying a Pipeline label in the Pipeline run details view for the enterprise contract pipeline. It changed from pipeline name to `enterprise-contract`.


Katka92:
[test]

PR #971, opened 2024-08-01
fix(ec): use only results from `report-json` step

zregvart:
## Fixes 

https://issues.redhat.com/browse/EC-562

## Description

This removes the logic to read YAML from the `report` step and relies on the JSON data present in the `report-json` step. We're considering removing the YAML output, so this is in preparation for that.

## Type of change

- [ ] Feature
- [ ] Bugfix
- [X] Refactoring (no functional changes, no api changes)
- [ ] Build related changes

## Browser conformance: 

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


src/components/EnterpriseContractView/useEnterpriseContractResultFromLogs.tsx
```diff
@@ -18,81 +15,39 @@ export const useEnterpriseContractResultFromLogs = (
...
-        .catch((err) => {
-          if (unmount) return;
-          if (err.code === 404) {
-            setFetchTknLogs(true);
-          } else {
-            setEcLoaded(true);
-          }
-          // eslint-disable-next-line no-console
-          console.warn('Error while fetching Enterprise Contract result from logs', err);
-        });
```
sahil143, line 58: This is required. When the pipeline run is running we fetch logs from Pod.
Below `getTaskRunLog` fetches the logs from tkn results.

sahil143 (changes requested):
@zregvart Thank you for the PR. One change is needed.

src/components/EnterpriseContractView/useEnterpriseContractResultFromLogs.tsx
zregvart: Am I wrong in my assumption that the logs are available on the "Logs" tab?

To me it seems that the `PipelineRunLogsTabProps` using `PipelineRunLogs` is responsible for that and this should only affect the data for the "Security" tab.

sahil143: Yes, logs are available only on the "Logs" tab. And for the security tab, it's the same mechanism as the logs tab if the pipeline is running, we use the pods to fetch the logs (i.e. results yaml) and parse the data for the security tab.
We don't have to wait for the pipeline run to finish to show the data for the security tab.

zregvart: I guess this relies on parsing partial YAML? This is an odd requirement for me, the logs from the `step-report` will be available after the main bulk of the work is done and milliseconds apart from the logs of the `report-json`. I do not think there is much benefit to the end user in doing this.

sahil143: In that case, I agree, We can remove that the logic to fetch the pod logs

zregvart:
@sahil143 can you approve & lgtm this?

sahil143 (approved):
/lgtm

lcarva:
/assign @rohitkrai03

As per the bot's suggestion. Assigning to @rohitkrai03 for approval.

PR #979, opened 2024-08-29
fix(test): remove sbom-json-check

Katka92:
## Description
The advanced test path started to fail as `sbom-json-check` task was removed ([PR#1330](https://github.com/konflux-ci/build-definitions/pull/1330)). Removing the task from tests too.


Katka92:
[test]

PR #988, opened 2024-09-24
fix(KFLUXBUGS-1197): component relationship remove button

caugello:
## Fixes 
Missing remove button on component relationship modal

## Description
Fixes [KFLUXBUGS-1197](https://issues.redhat.com//browse/KFLUXBUGS-1197)

## Type of change

- [x] Bugfix

## Browser conformance: 

- [x ] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge

### DIFF
--- a/src/components/ComponentRelation/__tests__/utils.spec.ts
+++ b/src/components/ComponentRelation/__tests__/utils.spec.ts
@@ -5,37 +5,6 @@ import {
   transformNudgeData,
 } from '../utils';
 
-describe('transformNudgeData', () => {
-  it('should transform data', () => {
-    expect(
-      transformNudgeData([
-        { source: 'abcd', nudgeType: ComponentRelationNudgeType.NUDGES, target: ['b', 'c'] },
-      ]),
-    ).toEqual({ abcd: ['b', 'c'] });
-    expect(
-      transformNudgeData([
-        { source: 'abcd', nudgeType: ComponentRelationNudgeType.NUDGES, target: ['b', 'c'] },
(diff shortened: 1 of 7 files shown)


sahil143 (changes requested):
@caugello Thank you for the PR.

Alignment and color of  remove icon seems off,
image

should be like this
image


When a relationship is removed, a API request should also be sent to update the relationship for the removed component. In the above screenshot If I remove the second relationship, A request should be sent to update the component `go-devfile-sample-vuln`. I guess we should compare the OldData with the new data from form here https://github.com/caugello/hac-dev/blob/bd7d19099be9c39c220baa5289d6142ce1025d86/src/components/ComponentRelation/utils.ts#L73-L74

src/components/ComponentRelation/__tests__/utils.spec.ts
```diff
@@ -1,5 +1,71 @@
 import { ComponentRelationNudgeType } from '../type';
-import { componentRelationValidationSchema, transformNudgeData } from '../utils';
+import {
+  componentRelationValidationSchema,
+  computeNudgeDataChanges,
+  transformNudgeData,
+} from '../utils';
+
+describe('transformNudgeData', () => {
```
sahil143, line 8: unit tests for `transformNudgeData` already exist, check below Line 70. Can we remove these? or Are these different?

sahil143 (changes requested):

src/components/ComponentRelation/__tests__/utils.spec.ts
caugello: It's a dupe, removed it.

sahil143 (approved):
/lgtm
/approve

PR #993, opened 2024-09-27
fix(KFLUXBUGS-603): update usage of build trigger

CryptoRodeo:
## Fixes 
Fixes https://issues.redhat.com/browse/KFLUXBUGS-603

## Description
In the current implementation on the component details page, the "Build trigger" is displayed as a link, but upon clicking, it simply opens a tooltip, which can be confusing for users.

A simple text value along with information icon (clickable) to indicate additional details, such as "Automatic " would be less confusing.

**Other changes:**
Remove the Build Pipeline Plan and Build Trigger columns from the component activity page.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

Build trigger in Component Details page:
**Before:**
image

**After:**
image

Component activity page section:
**Before:**
image

**After:**
image

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


sahil143 (commented):
/ok-to-test

sahil143 (changes requested):
there is a build trigger column on components list view also that has the same interaction, we will need to fix that also.

image

CryptoRodeo:
@sahil143 I reached out to UX to confirm if that should be changed as well. The ticket only mentions the component details page so I was unsure about the components list page.

src/components/Components/ComponentBuildTrigger.tsx
```diff
@@ -4,31 +4,37 @@ import { PACState } from '../../hooks/usePACState';
 
 type Props = {
   pacState: PACState;
+  withoutTooltip?: boolean;
```
abhinandan13jan, line 7: Should we remove this? If we are planning to keep it same in both the places?

src/components/Components/ComponentBuildTrigger.tsx
```diff
@@ -4,31 +4,37 @@ import { PACState } from '../../hooks/usePACState';
...
+
+const ComponentBuildTrigger: React.FC<React.PropsWithChildren<Props>> = ({
+  pacState,
+  withoutTooltip = false,
+}) => {
+  const buildTriggerText = getTriggerType(pacState);
+  return withoutTooltip ? (
+    <span>{buildTriggerText}</span>
+  ) : (
+    wrapInButtonTooltip(buildTriggerText, pacState)
```
abhinandan13jan, line 36: this is becoming a component in itself. Would it be better to use the function for returning the `text` only?

CryptoRodeo:
This change is still under discussion, I will add more changes once the implementation details have been confirmed.

CryptoRodeo:
Converted back to draft, it appears that we are removing columns now rather than updating the build trigger description.

CryptoRodeo:
@sahil143 , @abhinandan13jan On top of updating the build trigger in the component details page the PMs also wanted a change in the component activity page, which involved removing two columns.

Please see [the slack thread](https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1728416729438139?thread_ts=1728407468.514049&cid=C02CTEB3MMF ) for more details.

src/components/Components/ComponentDetails/__tests__/ComponentBuildSettings.spec.tsx
```diff
@@ -0,0 +1,34 @@
+import usePACState, { PACState } from '../../../../hooks/usePACState';
+import { getHelpPopoverText, getBuildTriggerText } from '../tabs/ComponentBuildSettings';
+
+jest.mock('../../../../hooks/usePACState');
+jest.mock('../../../../../src/components/HelpPopover');
+
+const usePacStateMock = usePACState as jest.Mock;
```
sahil143, line 7: mocks shouldn't be required here since only utils are being tested.

sahil143 (changes requested):

src/components/Components/ComponentDetails/__tests__/ComponentBuildSettings.spec.tsx
CryptoRodeo: Mock removed

sahil143 (approved):
/lgtm
/approve

CryptoRodeo:
Need to update path tests now that two columns have been removed from the pipeline runs tab.

sahil143 (approved):
/lgtm

PR #994, opened 2024-09-30
feat(KFLUXBUGS-1540): filter on Applications

caugello:
## Fixes 
[KFLUXBUGS-1540](https://issues.redhat.com//browse/KFLUXBUGS-1540)

## Description
Application list can be filtered by name

## Type of change

- [X] Feature

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge

### DIFF
--- a/src/components/ApplicationListView/ApplicationListView.tsx
+++ b/src/components/ApplicationListView/ApplicationListView.tsx
@@ -6,18 +6,22 @@ import {
   EmptyStateBody,
   PageSection,
   PageSectionVariants,
+  SearchInput,
   Spinner,
   Toolbar,
   ToolbarContent,
   ToolbarItem,
 } from '@patternfly/react-core';
+import { debounce } from 'lodash-es';
 import { HACBS_FLAG } from '../../hacbs/hacbsFeatureFlag';
 import { useApplications } from '../../hooks/useApplications';
+import { useSearchParam } from '../../hooks/useSearchParam';
 import emptyStateImgUrl from '../../imgs/Application.svg';
 import imageUrl from '../../imgs/getting-started-illustration.svg';
 import { ApplicationModel, ComponentModel } from '../../models';
 import { Table } from '../../shared';
 import AppEmptyState from '../../shared/components/empty-state/AppEmptyState';
+import FilteredEmptyState from '../../shared/components/empty-state/FilteredEmptyState';
 import { ApplicationKind } from '../../types';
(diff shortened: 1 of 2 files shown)


src/components/ApplicationListView/ApplicationListView.tsx
```diff
@@ -36,13 +40,35 @@ const ApplicationListView: React.FC<React.PropsWithChildren<unknown>> = () => {
...
+  }, [nameFilter]);
 
   const [applications, loaded] = useApplications(namespace);
   applications?.sort(
     (app1, app2) =>
       +new Date(app2.metadata.creationTimestamp) - +new Date(app1.metadata.creationTimestamp),
   );
 
+  const filteredApplications = React.useMemo(() => {
+    return applications?.filter((app) => app.metadata.name.includes(nameFilter));
```
sahil143, line 59: If `app.spec.displayName` is available, filter list based on that else fallback on `app.metadata.name`.

caugello: Added fallback and test

sahil143 (changes requested):
@caugello Filter doesn't seem to be working for case-sensitive terms. Maybe we need to convert both nameFilter and displayName to lower case before comparing.

image
image

src/components/ApplicationListView/ApplicationListView.tsx
```diff
@@ -36,13 +40,40 @@ const ApplicationListView: React.FC<React.PropsWithChildren<unknown>> = () => {
   const applicationBreadcrumbs = useApplicationBreadcrumbs();
   const [canCreateApplication] = useAccessReviewForModel(ApplicationModel, 'create');
   const [canCreateComponent] = useAccessReviewForModel(ComponentModel, 'create');
+  const [nameFilter, setNameFilter] = useSearchParam('name', '');
+  const [onLoadName, setOnLoadName] = React.useState(nameFilter);
+
+  React.useEffect(() => {
```
abhinandan13jan, line 46: Why do we need a `nameFilter` and also an `onLoadName` , in this hook, the `onLoadName` keeps updating with the `nameFilter` doesn't it?

caugello: Seems redundant, we can only use the query params from namefilter. Removed it.

caugello:
@sahil143 @abhinandan13jan anything blocking to merge this ?

PR #1005, opened 2024-10-14
fix(KFLUXBUGS-1680):  filter records on creationTimestamp

caugello:
## Fixes 
[KFLUXBUGS-1680](https://issues.redhat.com//browse/KFLUXBUGS-1680)

## Description
Querying pipelineruns or records may include all component names making the query to long and throws an HTTP error 414 or 431. Component names were included in the filter to avoid a visual issue when recreating an application with the same name, previously ran pipelineruns are displayed.

The component filter array is removed and uses the record creationTimestamp filter to be greater than the application creationTimestamp.

## Type of change

- [ ] Feature
- [X] Bugfix
- [ ] Build related changes

## How to test or reproduce?
Create a large number of components with large names.

[ ] Test A 
Initial issue:
When the query is too long, we should hit an http error 414 or as I do, a 431 locally.

 [ ] Test B 
Once change applied,  the filter includes a creationTimestamp attribute and removed the component matchExpression filter.

**Test Configuration(s)**:
I have an application with 240 components, named `devfile-sample-other-very-long-meaningless-name-${n}`

## Browser conformance: 

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/hooks/useBuildPipelines.ts
+++ b/src/hooks/useBuildPipelines.ts
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { PipelineRunLabel, PipelineRunType } from '../consts/pipelinerun';
 import { PipelineRunGroupVersionKind } from '../models';
 import { PipelineRunKind } from '../types';
+import { useApplication } from './useApplications';
 import { usePipelineRuns } from './usePipelineRuns';
 import { GetNextPage } from './useTektonResults';
 
@@ -13,31 +14,23 @@ export const useBuildPipelines = (
   componentNames?: string[],
   limit?: number,
 ): [PipelineRunKind[], boolean, unknown, GetNextPage] => {
+  const [application, applicationLoaded] = useApplication(namespace, applicationName);
   const [pipelineRuns, loaded, plrError, getNextPage] = usePipelineRuns(
-    includeComponents && !componentNames?.length ? null : namespace,
+    !applicationLoaded && includeComponents && !componentNames?.length ? null : namespace,
     React.useMemo(
       () => ({
         selector: {
+          filterByCreationTimestampAfter: application?.metadata?.creationTimestamp,
           matchLabels: {
             [PipelineRunLabel.APPLICATION]: applicationName,
             [PipelineRunLabel.PIPELINE_TYPE]: PipelineRunType.BUILD,
           },
-          ...(includeComponents &&
-            componentNames?.length > 0 && {
-              matchExpressions: [
-                {
-                  key: PipelineRunLabel.COMPONENT,
-                  operator: 'In',
-                  values: componentNames,
-                },
-              ],
-            }),
           filterByCommit: commit ? commit : undefined,
         },
         // TODO: Add limit when filtering by component name AND only PLRs are returned: https://github.com/tektoncd/results/issues/620
         // limit,
       }),
-      [applicationName, includeComponents, componentNames, commit],
+      [applicationName, commit, application],
     ),
   );
 
--- a/src/hooks/useLatestBuildPipelines.ts
+++ b/src/hooks/useLatestBuildPipelines.ts
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { PipelineRunLabel, PipelineRunType } from '../consts/pipelinerun';
 import { PipelineRunKind } from '../types';
+import { useApplication } from './useApplications';
 import { usePipelineRuns } from './usePipelineRuns';
 
 export const useLatestBuildPipelines = (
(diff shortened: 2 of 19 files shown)


caugello:
@sahil143 PTAL if this would be a suitable approach.

sahil143:
> @sahil143 PTAL if this would be a suitable approach.

@caugello We added all components to the api query because earlier if user deleted an application, pipeline runs in that application were not removed so in some cases user created a new application with same name and they were seeing pipeline runs from deleted application. We need to check is this is still the case if It's not, we can stop sending component names in the API request otherwise we need to raise this with the backend team. Ideally we shouldn't need to send all of the component names in the api request.

caugello:
> > @sahil143 PTAL if this would be a suitable approach.
> 
> @caugello We added all components to the api query because earlier if user deleted an application, pipeline runs from that application were not removed so in some cases user created a new application with same name and they were seeing pipeline runs from deleted application. We need to check is this is still the case if It's not, we can stop sending component names in the API request otherwise we need to raise this with the backend team. Ideally we shouldn't need to send all of the component names in the api request, only application name should suffice.

This is still the case. Deleted an application, re-created it. Pipeline-runs history from previous application with same name still present

caugello:
Applied changes to filter on creationTimestamp and removed component names filters causing the issue.

src/utils/tekton-results.ts
```diff
@@ -167,8 +183,6 @@ export const selectorToFilter = (selector?: Selector) => {
       if (matchExpressions) {
         filter = AND(filter, expressionsToFilter(matchExpressions));
       }
-    } else {
-      filter = labelsToFilter(selector as MatchLabels);
```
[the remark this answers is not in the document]
caugello: This is overwriting the filter. If a selector expression doesn't include a matchlabel or matchexpressions it will always execute the else part which is setting the incorrect filter hence the change in test output as well.  `filterByName` should have a `startsWith` expression instead of a label key assertion.

CryptoRodeo:
Overall LGTM, but I'll wait for Sahil to chime in :smile:

src/hooks/useBuildPipelines.ts
```diff
@@ -22,22 +22,12 @@ export const useBuildPipelines = (
...
-          ...(includeComponents &&
-            componentNames?.length > 0 && {
-              matchExpressions: [
-                {
-                  key: PipelineRunLabel.COMPONENT,
-                  operator: 'In',
-                  values: componentNames,
-                },
-              ],
-            }),
```
sahil143, line 34: we should add the creation timestamp filter here also. This also has the potential of fetching the deleted pipeline runs

src/hooks/useLatestBuildPipelines.ts
```diff
@@ -29,16 +29,9 @@ export const useLatestBuildPipelines = (
             [PipelineRunLabel.APPLICATION]: applicationName,
             [PipelineRunLabel.PIPELINE_TYPE]: PipelineRunType.BUILD,
           },
-          matchExpressions: [
-            {
-              key: PipelineRunLabel.COMPONENT,
-              operator: 'In',
-              values: neededNames,
-            },
-          ],
```
sahil143, line 38: same as above

src/hooks/usePACStatesForComponents.ts
```diff
@@ -83,16 +83,11 @@ const usePACStatesForComponents = (components: ComponentKind[]): PacStatesForCom
             [PipelineRunLabel.COMMIT_EVENT_TYPE_LABEL]: PipelineRunEventType.PUSH,
           },
           matchExpressions: [
-            {
-              key: PipelineRunLabel.COMPONENT,
-              operator: 'In',
-              values: neededNames,
-            },
```
sahil143, line 90: same as above

src/hooks/usePipelineRuns.ts
```diff
@@ -274,22 +274,12 @@ export const usePipelineRunsForCommit = (
...
-          ...(componentsLoaded &&
-            componentNames?.length > 0 && {
-              matchExpressions: [
-                {
-                  key: PipelineRunLabel.COMPONENT,
-                  operator: 'In',
-                  values: componentNames,
-                },
-              ],
-            }),
```
sahil143, line 286: same as above

sahil143 (changes requested):

src/hooks/useBuildPipelines.ts
caugello: updated all where requested

sahil143 (approved):
/lgtm

PR #1010, opened 2024-10-21
fix(e2e): enable sbom test, fix missing pageobjects

Katka92:
## Description
Fix missing page objects for Atlas. Enable advanced test again, with SBOM check.


sahil143 (approved):
/lgtm

PR #1044, opened 2025-04-07
chore(KFLUXVNGD-245): add tekton-bundle-builder-oci-ta to UI

avi-biton:
Add tekton-bundle-builder-oci-ta pipeline to Konflux UI Konflux UI would support dynamic pipeline when it is online. 
The hardcode would be used as the backup when there is no dynamic pipeline config found.


avi-biton:
@sahil143 Can you please take a look?

avi-biton:
Closing this PR, as updating the build configmap is enough, no need for updates in the UI

